PR #15, opened 2017-08-06
Switch CI to Docker

junghans:
Part of votca/votca#23


junghans:
@JensWehner: Welcome @codecov-io.

PR #19, opened 2018-02-08
Adding elements.h file

JoshuaSBrown:
This file is being added to tools. It is meant to consolidate the atomtable in csg and the elements.h file from xtp. 

In moving the file it has also been updated to include more of the elements from the periodic table. And exceptions have been added for methods that are deprecated.


JoshuaSBrown:
@JensWehner It would probably be good if you took a look at this before it is merged.

include/votca/tools/elements.h
```diff
@@ -0,0 +1,629 @@
...
+        _VdWMK["P"]  = 1.8;
+        _VdWMK["S"]  = 1.75;
+        _VdWMK["Cl"] = 1.7;
+        _VdWMK["Ag"] = 2.0;
+    };
+
+};
+}}
+
+#endif	/* __VOTCA_XTP_ELEMENTS_H */
```
junghans: Tools!

include/votca/tools/elements.h
```diff
@@ -0,0 +1,629 @@
+/* 
+ *            Copyright 2009-2017 The VOTCA Development Team
```
junghans: 2018

include/votca/tools/elements.h
```diff
@@ -0,0 +1,629 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#ifndef __VOTCA_XTP_ELEMENTS_H
```
junghans: `XTP`?

include/votca/tools/elements.h
```diff
@@ -0,0 +1,629 @@
...
+    
+*/
+class Elements
+{
+public:   
+
+    Elements() { FillMaps(); };
+   ~Elements() { };
+
+    const double     &getVdWChelpG(std::string name ) const
```
junghans: Please add doxygen documentation!

JoshuaSBrown:
Travis keeps failing with this 
```
#########################################################################################################
#                                                                                                       #
# ERROR:                                                                                                #
# run_gromacs.sh: There is a nan in 'confout.gro', this seems to be wrong.                              #
# For details see the logfile /home/votca/votca/build/csg-tutorials/methanol-water/X_0.5/re/inverse.log #
#                                                                                                       #
#########################################################################################################```

But I don't see how it is related to the file I added?

**Comment by junghans** [comment]
Just do a re-trigger.

File: include/votca/tools/elements.h [hunk 5]
```diff
@@ -0,0 +1,655 @@
...
+#include <stdexcept>
+#include <boost/numeric/ublas/matrix.hpp>
+#include <boost/numeric/ublas/vector.hpp>
+#include <boost/numeric/ublas/io.hpp>
+#include <iostream>
+#include <fstream>
+
+namespace votca { namespace tools {
+
+namespace ub = boost::numeric::ublas;
```
MrTheodor: Why do you need to set this namespace?

include/votca/tools/elements.h
```diff
@@ -0,0 +1,655 @@
...
+using namespace std;
+
+/**
+    \brief information about an element
+ 
+    The Elements class stores properties of several elements in the periodic table includeing: mass, element name, element symbol, van der Waals radius, effective nuclear potential, element number as it appears in the periodic table.
```
MrTheodor: Please put some new lines here.

include/votca/tools/elements.h
```diff
@@ -0,0 +1,655 @@
...
+    }
+
+    /// This method is now deprecated but was originally used for determining 
+    /// the effective nuclear potential of an atom. There are now more 
+    /// appropriate methods for doing this in the basisset.h and aobasis.h file
+    /// in xtp. 
+    const double     &getNucCrgECP(std::string name)  const 
+    { 
+        throw invalid_argument("CrgECP map is deprecated");
+        return _NucCrgECP.at(name);
```
MrTheodor: Remove return statement, it will never happen.

include/votca/tools/elements.h
```diff
@@ -0,0 +1,655 @@
...
+
+#ifndef __VOTCA_TOOLS_ELEMENTS_H
+#define	__VOTCA_TOOLS_ELEMENTS_H
+
+#include <string>
+#include <map>
+#include <stdexcept>
+#include <boost/numeric/ublas/matrix.hpp>
+#include <boost/numeric/ublas/vector.hpp>
+#include <boost/numeric/ublas/io.hpp>
```
MrTheodor: Do you need all `boost/numeric/ublas` headers?

include/votca/tools/elements.h
JoshuaSBrown: Yes, but won't that trigger warning when travis runs?

MrTheodor: I think that there shouldn't be any warning as there is a `throw` statement.

include/votca/tools/elements.h
JoshuaSBrown: Prepare to be wowed!

JoshuaSBrown:
Prepare to be wowed!

src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+foreach(PROG test_elements)
+  file(GLOB ${PROG}_SOURCES ${PROG}*.cc)
+  add_executable(unit_${PROG} ${${PROG}_SOURCES})
+  target_link_libraries(unit_${PROG} votca_tools)
```
junghans: You can put these three lines inside the `if(ENABLE_TESTING)` block, because there is no point in building the executable without testing.

src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+
+  if(ENABLE_TESTING)
+    add_test(run_unit_${PROG} unit_${PROG})
```
junghans: Is the `run_` prefix needed?

src/tests/test_elements.cc
```diff
@@ -0,0 +1,84 @@
+
```
junghans: Copyright missing!

src/tests/test_elements.cc
```diff
@@ -0,0 +1,84 @@
...
+    } catch (...) {
+      test = true;
+    }
+    assert(test);
+  }
+
+  cout << "Testing: getMass" << endl;
+  {
+    Elements ele;
+    assert(ele.getMass("K") = 39.098);
```
junghans: `==` ?

src/tests/test_elements.cc
```diff
@@ -0,0 +1,84 @@
...
+#include <cassert>
+#include <exception>
+#include <iostream>
+#include <string>
+#include <votca/tools/elements.h>
+
+using namespace std;
+using namespace votca::tools;
+
+int main(void) {
```
junghans: Maybe this is a good point to use something like boost test framework?

JoshuaSBrown: Hmm I will have to look into it something else I don't know about. But more unit testing and pull requests will be coming your way so it's probably worth it.

src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+
+if(ENABLE_TESTING)
+  foreach(PROG test_elements )
+    ADD_DEFINITIONS(-DBOOST_TEST_DYN_LINK)
```
junghans: Will this break linking against static boost test framework libraries?

And please use:
```
target_compile_definitions(unit_${PROG} PRIVATE BOOST_TEST_DYN_LINK)
```
instead.

src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+
+if(ENABLE_TESTING)
+  foreach(PROG test_elements )
```
junghans: I am not sure, why you are doing this in a loop, but I might make it easier to add more tests later.

JoshuaSBrown: Yes, that was my reasoning.

src/tests/CMakeLists.txt
JoshuaSBrown: I'm not sure. Are we using static libraries for boost in general, or is it meant to be an option where both are supported? If so is there a variable associated with static vs dynamic linking defined somewhere in the cmake files?

junghans: I was just curious. The main problem being that distribution don't package static libraries, so let's leave like this for now and wait for complains.

junghans:
Looking at https://codecov.io/gh/votca/tools/pull/19, I think we should exclude the `src/tests/*` from the coverage report.

Can you copy this file https://github.com/junghans/travis-docker-glue/blob/master/.codecov.yml?

JoshuaSBrown:
Are you sure that's kind of the point of testing right? The reason nothing is showing up but the elements.h is because it is the only thing with actual run time testing.

CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ endif(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/.git)
 find_package(Threads REQUIRED)
 set(THREAD_LIBRARIES ${CMAKE_THREAD_LIBS_INIT})
 
-find_package(Boost 1.39.0 REQUIRED COMPONENTS program_options filesystem system )
+find_package(Boost 1.39.0 REQUIRED COMPONENTS program_options unit_test_framework filesystem system )
```
junghans: I think we should only look for `unit_test_framework` when testing is enabled.

src/CMakeLists.txt
```diff
@@ -1,2 +1,3 @@
 add_subdirectory(libtools)
 add_subdirectory(tools)
+add_subdirectory(tests)
```
junghans: We could put the `if(ENABLE_TESTING)` around the `add_subdirectory(tests)`, then we don't have to do it in the `CMakeLists.txt` itself.

src/tests/.clang-format
```diff
@@ -0,0 +1,49 @@
+---
```
junghans: Can we move this file to the root of the repository.

src/tests/test_elements.cc
junghans: Fixed!

CMakeLists.txt
JoshuaSBrown: Good idea.

src/tests/.clang-format
JoshuaSBrown: So I don't know that much about clang-format do you know if it will pick up on .clang-format if it is not in the cwd?

junghans:
> Are you sure that's kind of the point of testing right? The reason nothing is showing up but the elements.h is because it is the only thing with actual run time testing.

Yes, the test itself should not go into the coverage number and hence being excluded. And the link above shows the coverage diff. The coverage of master (with this PR) is here: https://codecov.io/gh/votca/tools/branch/master

CMakeLists.txt
junghans: There is one other problem. `libvotca_tools` should be linked against `libboost_unit_test_framework` as it doesn't use any symbol from that library! This is to prevent over-linking.

So we need to split `libboost_unit_test_framework` out of `Boost_LIBRARIES`, which `libvotca_tools` is linked against.

src/tests/.clang-format
junghans: It will go through all parent directories until it find one.

CMakeLists.txt
junghans: Ok, see the commit is just pushed.

JoshuaSBrown: Thanks for doing that I would have had a hard time figuring that out.

JoshuaSBrown:
I think I have addressed everything... I'm sure you will let me know if I missed something.

junghans:
Now as merged, don’t forget to remove atom tables from csg and xtp!

PR #21, opened 2018-02-12
Edge class

JoshuaSBrown:
Edge Class is a general class for describing the relationship between two vertices or points. It is more specific then just a pair class, it makes use of underlying pair class
(p1,p2)
1. Stores two int values 
2. Smaller int is always put in p1 location
3. Allows sorting and hashes 
 
Useful for describing graphs, bonds, links molecular structures etc....

### DIFF
--- a/include/votca/tools/edge.h
+++ b/include/votca/tools/edge.h
@@ -1,0 +1,91 @@
+/*
+ *            Copyright 2018-2018 The VOTCA Development Team
+ *                       (http://www.votca.org)
+ *
+ *      Licensed under the Apache License, Version 2.0 (the "License")
+ *
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *              http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include <iostream>
+#include <utility>
+
+#ifndef _VOTCA_TOOLS_EDGE_H
+#define _VOTCA_TOOLS_EDGE_H
+
+namespace votca {
+namespace tools {
+
+/**
+ * \brief Connects to vertices
+ *
+ * The edge class stores the ids of two seperate vertices indictating that they
+ * are connected (id1,id2). Unlike a pair the vertex with the lower value is
+ * always placed in id1, this allows us to reduce ambiguity when dealing with
+ * a link.
+ */
+class Edge {
+ private:
+  std::pair<int, int> vertices;
+
+ public:
+  Edge() {}
+  ~Edge() {}
+  /// Creates an edge the smallest integer value will be placed in the id1
+  /// spot and the larger in the id2 spot
+  Edge(int ID1, int ID2);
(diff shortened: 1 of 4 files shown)


JoshuaSBrown:
I will add one now I was waiting for the elements.h file to be merged.

include/votca/tools/edge.h
```diff
@@ -0,0 +1,91 @@
+/*
+ *            Copyright 2018-2018 The VOTCA Development Team
```
MrTheodor: 2018-2018 ?

include/votca/tools/edge.h
```diff
@@ -0,0 +1,91 @@
...
+  Edge() {}
+  ~Edge() {}
+  /// Creates an edge the smallest integer value will be placed in the id1
+  /// spot and the larger in the id2 spot
+  Edge(int ID1, int ID2);
+  Edge(const Edge& ed) : vertices(ed.vertices) {}
+  /// Given one of the integers in the edge the other will be output
+  int getOtherV(int ver) const;
+  /// grab the smaller integer
+  int getV1(void) const { return vertices.first; }
```
MrTheodor: in C++ there is no difference between `getV1()` and `getV1(void)`

src/tests/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 find_package(Boost 1.39.0 REQUIRED COMPONENTS unit_test_framework)
-foreach(PROG test_elements )
+foreach(PROG test_elements test_edge )
```
MrTheodor: Why not use `file(GLOB test_cc test_*.cc)` to get the list of tests; you can use `get_filename_component(fname ${PROG} NAME_WE)` to get the file name without the extension of the directory.

https://cmake.org/cmake/help/v3.0/command/get_filename_component.html

src/tests/test_edge.cc
```diff
@@ -0,0 +1,89 @@
...
+  cout << ed << endl;
+}
+
+BOOST_AUTO_TEST_CASE(equivalence_test) {
+  Edge ed(2, 3);
+  Edge ed2(2, 4);
+  Edge ed3(3, 2);
+  BOOST_CHECK_EQUAL(ed, ed);
+  BOOST_CHECK_EQUAL(ed, ed3);
+  BOOST_CHECK_EQUAL((ed == ed2), 0);
```
MrTheodor: please, change `0` to `false` as you did in the next test_case

src/tests/CMakeLists.txt
JoshuaSBrown: That seems better, I'm just not yet that familiar with cmake will definitely implement.

JoshuaSBrown:
Let me make sure what I want to do is not already done by boost before this is merged.

src/tests/CMakeLists.txt
junghans: I would prefer listing the tests explicitly in case we have to disable a test.

I would use something like
```cmake
foreach(TEST elements edge)
```
though.

JoshuaSBrown:
I have parsed over the boost graph library - briefly so I might have missed something.  
1. Non of the algorithms it uses do what I would like to do, so I would have to write it from scratch anyway. 
2. Regardless the edge class provides an easy means to compare a bond (undirected edge) as it does not depend on what order you put the vertices into the constructor. 
3. At it's core it is a pair but with extra functionality, whereas the boost library simply uses pairs.

src/tests/CMakeLists.txt
JoshuaSBrown: Ok I will leave it then :)

PR #22, opened 2018-02-12
Type Converter class

JoshuaSBrown:
Added type converter class provides the ability to convert between complex types without the need for cross linking multiple header files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,18 @@ services:
 
 env:
   matrix:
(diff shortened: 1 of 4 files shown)


JoshuaSBrown:
I will update the CMakeLists.txt file in tools/tests after the edge class is merged.

JoshuaSBrown:
Some of our repos... I won't say which one could benefit from a reduction in include statements in the header files. If must conversions are handled through this object there is no need to to include multiple functions simply responsible for converting from one type to another. I realize the object is a little abstract maybe as I use it the benefits will become more apparent.

PR #26, opened 2018-02-15
1. Added identity object

JoshuaSBrown:
2. Added test for identity object
3. Updated CMakeLists.txt


JoshuaSBrown:
In hind sight we should have a discussions about this before it gets merged.

PR #28, opened 2018-02-19
Generic graph node

JoshuaSBrown:
This is the basic node object which will be used with the graph like algorithms. It also serves as a unique identifier for the contents of an object. Ultimately it is designed to help with the mapping from molecules to rigid fragments and segments in the xtp repo. However, functionality could be easily extended to csg if desired.


JoshuaSBrown:
Your talking about this file: https://github.com/votca/csg/blob/master/src/libcsg/map.cc

junghans:
Yeah!

JoshuaSBrown:
Ok, will keep it in mind, that code considers shapes such as spheres and ellipsoidal shapes to determine neighbors etc?

junghans:
Yes, spherical mapping is the default, but non-spherical mapping are implemented, too!

JoshuaSBrown:
@MrTheodor  have you ever used this: http://www.boost.org/doc/libs/1_55_0/libs/property_map/doc/property_map.html
I'm wondering if the boost property map has similar functionality to my graph node.

junghans:
Can we build this on top of property map?

JoshuaSBrown:
I can try.

JoshuaSBrown:
I just saw this which makes me think writing part of the graph library for what I need is probably not that uncommon even with boost: 
```The default method of associating a value with an object is to make the value a member. I.e. struct Edge { float value; }. If you need more that one value, provide your own struct. Nothing really different from struct RGB { int R; int G; int B; }; std::vector<RGB>. BGL is so complex that I think you should roll your own in 90% of cases, using BGL only for the hardest 10%. Compare that to STL, which is sufficient for the simplest 90% of cases - the total opposite. – MSalters```

JoshuaSBrown:
Turns out the property map is not doing the same thing that I want to do but I will keep my eyes open for functionality that is available in boost so we are not recreating code.

PR #29, opened 2018-02-20
Edge container

JoshuaSBrown:
Used to store groups of edges and provides standard operations that might be useful for instance in a graph such as finding neighboring vertices.


src/tests/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 find_package(Boost 1.39.0 REQUIRED COMPONENTS unit_test_framework)
-foreach(PROG test_elements test_edge test_name test_typeconverter )
-  file(GLOB ${PROG}_SOURCES ${PROG}*.cc)
+foreach(PROG test_elements test_edge test_name test_typeconverter test_graphnode test_edgecontainer)
+  file(GLOB ${PROG}_SOURCES ${PROG}.cc)
```
junghans: If you drop the `*` the whole globbing command can dropped! It would also disallow for a test using multiple files!

include/votca/tools/graphnode.h
```diff
@@ -0,0 +1,65 @@
+/*
+ *            Copyright 2018-2018 The VOTCA Development Team
```
junghans: 2018-2018?

src/tests/CMakeLists.txt
junghans: It might be better to rename `test_edge` to `test_edge_simple`.

JoshuaSBrown: Fair enough, but why do we need the GLOB anyway if we are going to be explicitly writing out all the tests?

src/libtools/edgecontainer.cc
```diff
@@ -0,0 +1,87 @@
...
+    f_verts.insert(ed.getV2());
+    f_eds.insert(ed);
+  }
+
+  for (auto vert : f_verts) vertices_.push_back(vert);
+  for (auto ed : f_eds) edges_.push_back(ed);
+}
+
+bool EdgeContainer::edgeExist(Edge ed) {
+  return (find(edges_.begin(), edges_.end(), ed) != edges_.end());
```
MrTheodor: How often will you ask if an edge exists and how often you will add/remove edges? vector::find complexity is O(N)

src/libtools/edgecontainer.cc
```diff
@@ -0,0 +1,87 @@
...
+void EdgeContainer::addEdge(Edge ed) {
+  if (!edgeExist(ed)) edges_.push_back(ed);
+  if (!vertexExist(ed.getV1())) vertices_.push_back(ed.getV1());
+  if (!vertexExist(ed.getV2())) vertices_.push_back(ed.getV2());
+  return;
+}
+
+vector<int> EdgeContainer::getVertices() { return vertices_; }
+
+vector<int> EdgeContainer::getNeighVertices(int vert) {
```
MrTheodor: Please have a look at adjacency list concept. This could really simplify such queries.

src/libtools/edgecontainer.cc
JoshuaSBrown: Quite often. Do you have a better solution? I have not implemented a remove edges method yet. Will probably have to in the future though.

src/libtools/edgecontainer.cc
JoshuaSBrown: Good point, I'm going to have to refactor perhaps I can pull in some of the functionality from boost and use my functions as wrappers.

src/libtools/edgecontainer.cc
JoshuaSBrown: Though the remove Edge option is not likely to be used much.

src/tests/CMakeLists.txt
junghans: Well you copied the loop from `csg`. And there the idea was that all files which start with a certain pattern get compiled into one executable, e.g. `csg_stat*.cc` matches `csg_stat.cc` and `csg_stat_imc.cc` and both files get compiled into the `csg_stat` executable. Downside is of course, that the program prefix has to be unique, which `test_edge` isn't, when having files like `test_edge_container.cc`.

I guess the globbing would make sense here, too, if you don't want to put all e.g. edge-related test in one file.

PR #34, opened 2018-03-22
1. Removed overload == and != operators for matrix from header and te…

JoshuaSBrown:
…st file

2. Replaced with isClose function that takes the tolerance as a parameter
3. Added a floating point comparison function to a new file and added a test for it

### DIFF
--- a/include/votca/tools/matrix.h
+++ b/include/votca/tools/matrix.h
@@ -63,7 +63,7 @@ namespace tools {
        * @param[in] tol - tolerance 
        * @return bool - return true if within tolerance and false if not
        */
-      bool isClose(const matrix m, double tol) const {
+      bool isClose(const matrix& m, double tol) const {
         for (size_t i = 0; i < 9; ++i){
           if (!isApproximatelyEqual(_m[i], m._m[i],tol)) return false;
         }
--- a/include/votca/tools/matrix.h
+++ b/include/votca/tools/matrix.h
@@ -18,9 +18,9 @@
 #ifndef _VOTCA_TOOLS_MAT_H
 #define _VOTCA_TOOLS_MAT_H
 #include "ostream"
-#include <votca/tools/floatingpointcomparison.h>
-#include <votca/tools/types.h>
-#include <votca/tools/vec.h>
+#include "floatingpointcomparison.h"
+#include "types.h"
+#include "vec.h"
 
 namespace votca {
 namespace tools {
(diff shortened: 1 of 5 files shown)


JoshuaSBrown:
@JensWehner check if you think this is better.

include/votca/tools/matrix.h
```diff
@@ -17,10 +17,10 @@
 
 #ifndef _VOTCA_TOOLS_MAT_H
 #define _VOTCA_TOOLS_MAT_H
-
 #include "ostream"
-#include "types.h"
-#include "vec.h"
+#include <votca/tools/floatingpointcomparison.h>
+#include <votca/tools/types.h>
+#include <votca/tools/vec.h>
```
junghans: For headers in the same package, please use relative path (as it was before for vec.h)

JoshuaSBrown: Is this good practice or just because its been the convention?

junghans: Convention and consistency with other headers!

junghans: https://stackoverflow.com/questions/597318/what-are-the-benefits-of-a-relative-path-such-as-include-header-h-for-a-hea

I don't think it matters much, but consistency!

JoshuaSBrown:
@JensWehner  Are you saying replace 
```inline bool operator==(const vec &v1, const vec &v2)
{
    return ((v1.getX()==v2.getX()) && (v1.getY()==v2.getY()) && (v1.getZ()==v2.getZ()));
}

inline bool operator!=(const vec &v1, const vec &v2)
{
    return ((v1.getX()!=v2.getX()) || (v1.getY()!=v2.getY()) || (v1.getZ()==v2.getZ()));
}

```
with a isClose function?

JoshuaSBrown:
@JensWehner yes very nice! You  still didn't answer my question though.

JensWehner:
What was your question?

JensWehner:
yeah we should replace them, @onurcaylak  is working on that.

JoshuaSBrown:
Should I leave this branch open for that or do you want to add that to this one?

JensWehner:
you can close after merging.

Good job

JensWehner:
@junghans can you force merge? I do not know why travis fails?

junghans:
@JensWehner some API in gromacs master, should be fixed in votca/csg#244.

PR #37, opened 2018-04-09
Vector close feature

onurcaylak:
Added isClose to vec class,
added unittest for vec
fixed Warning in elements.h
prepared all tools dependencies of vec.h for removal of using namespace std: in vec.h. Retained because ctp fails if using namespace is removed

### DIFF
--- a/include/votca/tools/elements.h
+++ b/include/votca/tools/elements.h
@@ -84,7 +84,7 @@ class Elements {
   /// Returns the mass of each atom in a.u.
   const double &getMass(std::string name) const { return _Mass.at(name); }
   /// Returns the covalent Radii of the atom 
-  const double getCovRad(std::string name,std::string unit ) const 
+  double getCovRad(std::string name,std::string unit ) const 
   { 
     //TODO - This should be replaced with an object, an object that should
     //       auto recognise the units and return it in a standard type
--- a/include/votca/tools/vec.h
+++ b/include/votca/tools/vec.h
@@ -104,7 +104,22 @@ public:
      */
     const double &getZ() const { return _z; }
     
-    
+     /**
+       * \brief Compare floating point values of two vectors
+       *
+       * The floating point values are compared to within a tolerance which
+       * is specified as the third parameter.
+       *
+       * @param[in] v - vector to be compared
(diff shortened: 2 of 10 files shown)


include/votca/tools/vec.h
```diff
@@ -104,7 +105,22 @@ class vec {
...
+       * @param[in] v - vector to be compared
+       * @param[in] tol - tolerance 
+       * @return bool - return true if within tolerance and false if not
+       */
+      bool isClose(const vec& v, double tol) const {
+        for (size_t i = 0; i < 3; ++i){
+          if (!isApproximatelyEqual(xyz[i], v.xyz[i],tol)) return false;
+        }
+        return true;
+}
```
JoshuaSBrown: You can the inline command to this function as well.

JoshuaSBrown: Make tol a const double as well .

include/votca/tools/vec.h
```diff
@@ -104,7 +105,22 @@ class vec {
...
-    
+     /**
+       * \brief Compare floating point values of two vectors
+       *
+       * The floating point values are compared to within a tolerance which
+       * is specified as the third parameter.
+       *
+       * @param[in] v - vector to be compared
+       * @param[in] tol - tolerance 
+       * @return bool - return true if within tolerance and false if not
```
JoshuaSBrown: nice

include/votca/tools/elements.h
```diff
@@ -84,7 +84,7 @@ class Elements {
   /// Returns the mass of each atom in a.u.
   const double &getMass(std::string name) const { return _Mass.at(name); }
   /// Returns the covalent Radii of the atom 
-  const double getCovRad(std::string name,std::string unit ) const 
+  double getCovRad(std::string name,std::string unit ) const
```
junghans: Why?

JensWehner: gives compiler warnings, because you return a variable by copy but const declare it

include/votca/tools/vec.h
JensWehner: yeah inlining, I would prefer to have .cc files soon.

PR #38, opened 2018-04-09
test_matrix: fix warning

junghans:


JoshuaSBrown:
What was the warning?

JoshuaSBrown:
Instead of deleting can you just print it:
`cout << mat4 << endl;`

junghans:
Good point, can you do a pull request to `warning` branch?

PR #42, opened 2018-06-08
Qr solve fix

JensWehner:
Restructured qr_solve_constrained a bit
removed unnecssary minus sign
added unit test

this breaks CSG a bit, see pull request in CSG


junghans:
@schererc: Can you please check that this cleanup doesn't alter your `csg_fmatch` results?

JensWehner:
Before it did not give x but -x as the solution.

junghans:
@JensWehner might a force vs dV/dr thing.

junghans:
Oh well, I meant test in combination with votca/csg#248.

JensWehner:
@junghans Good point, but I think it is not because there were a couple of comments in the code, saying ` b = -y; // why is it -y?`
and
`// TODO: here i changed the sign, check again! (victor)\
x = -prec_prod( Q, x );`

junghans:
I think, you removed one minus in tools and one in csg, so we should be good.

PR #44, opened 2018-06-16
Generic graph

JoshuaSBrown:
I have completed the first draft of the generic graph classes. These classes will form the backbone of development into csg as I update and enhance the topological mapping algorithms by providing automation.


src/tests/CMakeLists.txt
```diff
@@ -3,7 +3,12 @@ list(APPEND test_cases test_edge_base)
 list(APPEND test_cases test_edgecontainer)
 list(APPEND test_cases test_elements)
 list(APPEND test_cases test_floatingpointcomparison)
+list(APPEND test_cases test_graph_base)
+list(APPEND test_cases test_graph_bf_visitor)
+list(APPEND test_cases test_graphalgorithm)
+list(APPEND test_cases test_graphdistvisitor)
 list(APPEND test_cases test_graphnode)
+list(APPEND test_cases test_graphvisitor)
```
junghans: Huh, why are you doing this list append? I think all tests should just be listed in the `foreach()` statement.

JoshuaSBrown: Because that turns into a really long single line.

JoshuaSBrown: Unless you know of a way to break up such lines?

junghans: Just do a line break, e.g.:
```cmake
foreach(i a b c
	d e f)
  message("${i}")
endforeach()
```

JoshuaSBrown: Well that's easy enough 👍

JoshuaSBrown:
@MrTheodor @pmadhikar if either of you have time could you look over this before I merge it?

include/votca/tools/edgecontainer.h
```diff
@@ -35,14 +36,18 @@ namespace tools {
...
+ protected:
   std::unordered_map<int,std::set<int>> adj_list_; 
  public:
   /// Constructors can take no arguments a single Edge or a vector of edges
   EdgeContainer() {};
   EdgeContainer(Edge ed);
   EdgeContainer(std::vector<Edge> eds);
 
+  /// Get the value of the max degree
+  int getMaxDegree(void);
```
MrTheodor: in C++ you do not need `(void)`

include/votca/tools/graph.h
```diff
@@ -0,0 +1,126 @@
...
+
+class GraphNode;
+
+class Graph : public EdgeContainer {
+ private:
+  /// Parameter description 
+  /// @param int - is the index of the graph nodes / vertex ids
+  /// @param GraphNode - this is the node object at each vertex and contains 
+  /// all the informatino that is relevant to that object
+  std::unordered_map<int, GraphNode> nodes_;
```
MrTheodor: Actually, you could also think about `multimap`, see here:
https://www.fluentcpp.com/2018/04/10/maps-vectors-multimap/

include/votca/tools/graph.h
```diff
@@ -0,0 +1,126 @@
...
+
+  /// This is the id of the graph to graphs that contain the same content 
+  /// are considered equal
+  std::string id_;
+
+  /// If the id has been calculated yet
+  bool id_set_;
+  /// Update the graphs ids
+  void updateIds_(Graph& g);
```
MrTheodor: what does this method do? is it updating `this` graph id or it is a static method that takes any Graph and updates `id` ?

include/votca/tools/graph.h
```diff
@@ -0,0 +1,126 @@
...
+  /// vertex) and return them in a vector with their corresponding graph node.
+  std::vector<std::pair<int, GraphNode>> getIsolatedNodes(void);
+
+  /// Functions determines which vertices do not have a graph node associated
+  /// with them and return their ids in a vector.
+  std::vector<int> getVerticesMissingNodes(void);
+
+  /// Returns a vector of the vertices and their graph nodes that are directly
+  /// connected to the vertex 'vert'
+  std::vector<std::pair<int, GraphNode>> getNeighNodes(int vert);
```
MrTheodor: is it worth to save four letters? ;-)

include/votca/tools/graph.h
```diff
@@ -0,0 +1,126 @@
...
+ * contetns of the graph node object are used to determine precidence e.g.
+ *
+ * pair<int,GraphNode> pr_grn1{ 1, gn };
+ * pair<int,GraphNode> pr_grn2{ 2, gn2 };
+ *
+ * vector<pair<int,GraphNode> > vec_pr_gn = { pr_grn1, pr_grn2 , ... etc };
+ *
+ * sort(vec_pr_gn.begin(),vec_pr_gn.end(),cmpVertNodePair);
+ */
+bool cmpVertNodePair(std::pair<int, GraphNode> gn1_pr,
```
MrTheodor: or two letters ;-)

include/votca/tools/graphalgorithm.h
```diff
@@ -0,0 +1,106 @@
...
+  auto verts = g.getVerticesDegree(maxD);
+
+  // Get the nodes and determine which node has the greatest stringID
+  // When compared using compare function
+  std::string str_id = "";
+  std::vector<int> gn_ids;
+  for (auto v : verts) {
+    auto gn = g.getNode(v);
+    int comp_int = str_id.compare(gn.getStringId());
+    if (comp_int > 0) {
```
MrTheodor: you could simply write `if (str_id.compare(gn.getStringId()))`

btw, why `id` is a string and not an integer ?

src/libtools/edgecontainer.cc
```diff
@@ -37,11 +37,29 @@ EdgeContainer::EdgeContainer(vector<Edge> eds) {
   }
 }
 
+int EdgeContainer::getMaxDegree(void){
+  int max = 0;
+  for(auto const& it : adj_list_) {
+    if(lexical_cast<int>(it.second.size())>max) max = it.second.size();
```
MrTheodor: why? what the `lexical_cast` is doing here?

src/libtools/graph.cc
```diff
@@ -0,0 +1,129 @@
...
+#include <votca/tools/graph.h>
+
+using namespace std;
+
+namespace votca {
+namespace tools {
+
+class GraphNode;
+
+void Graph::updateIds_(Graph& g) {
```
MrTheodor: ?? why this method updates `id_` of `this` object and do the update of external `g` object?

src/libtools/graph.cc
```diff
@@ -0,0 +1,129 @@
...
+    this->calcId_();
+    this->id_set_ = true;
+  }
+  if (!g.id_set_) {
+    g.calcId_();
+    g.id_set_ = true;
+  }
+}
+
+bool Graph::operator!=(Graph& g) {
```
MrTheodor: I would never expect that `compare` operator will update the objects..

src/libtools/graphalgorithm.cc
```diff
@@ -0,0 +1,52 @@
...
+namespace votca {
+namespace tools {
+
+bool singleNetwork(Graph g, GraphVisitor& gv){
+
+  exploreGraph(g,gv);
+  auto all_vertices = g.getVertices();
+  auto exp_vertices = gv.getExploredVertices(); 
+  auto iso_nodes = g.getIsolatedNodes();
+  return exp_vertices.size()==all_vertices.size() &&
```
MrTheodor: put it in a single line

src/libtools/graphalgorithm.cc
```diff
@@ -0,0 +1,52 @@
...
+  return exp_vertices.size()==all_vertices.size() && 
+         iso_nodes.size() == 0;
+}
+
+void exploreGraph(Graph& g, GraphVisitor& gv, int starting_vertex){
+
+  // Create a list of all vertices and determine if they have all been
+  // explored when the visitor que is empty 
+  gv.startingVertex(g,starting_vertex);
+  while(gv.queEmpty()==false){
```
MrTheodor: `while(!gv.queEmpty())`

src/libtools/graphnode.cc
```diff
@@ -133,12 +133,65 @@ void GraphNode::setStr(const unordered_map<string, string> str_vals) {
...
 
+int GraphNode::getInt(const string str) {
+  if (int_vals_.count(str) == 0)
+    throw invalid_argument(
+        "GraphNode does not "
+        "contain value");
+  return int_vals_[str];
+}
+
+int GraphNode::getDouble(const string str) {
```
MrTheodor: shouldn't the return type be `double` ?

src/libtools/graphnode.cc
```diff
@@ -133,12 +133,65 @@ void GraphNode::setStr(const unordered_map<string, string> str_vals) {
...
+  if (double_vals_.count(str) == 0)
+    throw invalid_argument(
+        "GraphNode does not "
+        "contain value");
+  return double_vals_[str];
+}
+
+int GraphNode::getStr(const string str) {
```
MrTheodor: and here `string`

src/libtools/graphnode.cc
```diff
@@ -133,12 +133,65 @@ void GraphNode::setStr(const unordered_map<string, string> str_vals) {
...
+  if (str_vals_.count(str) == 0)
+    throw invalid_argument(
+        "GraphNode does not "
+        "contain value");
+  return double_vals_[str];
```
MrTheodor: `return str_vals_[str]` ?

src/tests/test_graph_base.cc
```diff
@@ -0,0 +1,292 @@
...
+  v = round(v);
+  v /= pow(10, p);
+  return v;
+}
+
+BOOST_AUTO_TEST_SUITE(graph_test)
+
+BOOST_AUTO_TEST_CASE(constructors_test) { Graph g; }
+
+BOOST_AUTO_TEST_CASE(isolatednodes_test) {
```
MrTheodor: please add some doc with information what do you plan to test in the unit

pmadhikar:
I'm away until next week so I won't be able to review for a while. But it looks like the review is in more capable hands anyway :)

include/votca/tools/edge.h
```diff
@@ -75,6 +76,10 @@ class Edge {
   /// Print the contents of the edge
   friend std::ostream& operator<<(std::ostream& os, const Edge ed);
 };
+
+// Value used as a dummy object
+const Edge DUMMY_EDGE(std::numeric_limits<int>::max(),
```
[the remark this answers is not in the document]
JoshuaSBrown: Ah that must have been changed when I ran the google formatting style on the file. Good spot.

include/votca/tools/graph.h
JoshuaSBrown: Ya, I guess it is not very clear, it is used before the graphs are compared with "!=" and "==" the id is what is used to determine if the graphs are equal. The id is a string that is composed based on the sorted contents of each of the graph nodes. Thus two graphs will be equal if their graph nodes are the same. Because it is costly to create the id it is made once and only updated if a graph node has been changed or if it has not yet been set.

JoshuaSBrown: I think perhaps it would be clearer if it only updated it's own id.

src/libtools/graph.cc
JoshuaSBrown: It is not really updating the object in the sense that it is changing what's in the graph, it is only creating an id that summarizes the contents of the graph in a string. It is easier to compare a string than to compare all the contents each time I want to see if two graphs are equal. The id itself is private so the user is not supposed to know or use it.

src/libtools/graph.cc
JoshuaSBrown: Good point I have corrected this because it is confusing.

src/libtools/graphalgorithm.cc
JoshuaSBrown: ok

src/libtools/graphnode.cc
JoshuaSBrown: :) Oh ya good catch.

include/votca/tools/graph.h
JoshuaSBrown: Yes :)

src/tests/test_graph_base.cc
JoshuaSBrown: Good point.

include/votca/tools/graph.h
MrTheodor: Just a tip :)

src/libtools/graph.cc
MrTheodor: Actually, if you look into the docs (https://en.cppreference.com/w/cpp/language/operator_comparison) this kind of operator should be `const`. I am simply worried about the unexpected side effect.

include/votca/tools/graph.h
MrTheodor: Ok ;)

src/libtools/edgecontainer.cc
JoshuaSBrown: I guess this explains why I had the lexical cast, size() returns type `size_t` whereas max is of type int:

[  2%] Building CXX object tools/src/libtools/CMakeFiles/votca_tools.dir/graph_bf_visitor.cc.o
/home/votca/votca/tools/src/libtools/edgecontainer.cc: In member function 'int votca::tools::EdgeContainer::getMaxDegree()':
/home/votca/votca/tools/src/libtools/edgecontainer.cc:43:24: error: comparison of integer expressions of different signedness: 'std::set<int>::size_type' {aka 'long unsigned int'} and 'int' [-Werror=sign-compare]
     if(it.second.size()>max) max = it.second.size();
        ~~~~~~~~~~~~~~~~^~~~
cc1plus: all warnings being treated as errors
make[2]: *** [tools/src/libtools/CMakeFiles/votca_tools.dir/build.make:167: tools/src/libtools/CMakeFiles/votca_tools.dir/edgecontainer.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:244: tools/src/libtools/CMakeFiles/votca_tools.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
The command '/bin/sh -c make -j2' returned a non-zero code: 2

src/libtools/graph.cc
JoshuaSBrown: Oh I see, I think I can make it constant. If the id is always up to date, then there is no reason to call that update function when the equivalence operator is called. Good point!

JoshuaSBrown: Wow you just helped me find a design flaw in this class :) thanks.

PR #46, opened 2018-07-05
removed random class

JensWehner:
removed the random class 
as the random2 class exists and is used everywhere.


junghans:
I think it would be good to switch `Random2`to Mersenne Twister, which comes with C++11.

JensWehner:
I did some research on the issue and that is not a good idea. The algorithm currently implemented in tools is a lot faster than Mersenne Twister, whilst giving approximately the same quality of randomness.

JensWehner:
A also did a benchmark on it and at east the speed part is true.

junghans:
Most important question, is it implemented correctly?

JensWehner:
I do not know.

PR #48, opened 2018-07-06
rewrote random2 a bit to be more c++ and remove macros and unused fun…

JensWehner:
…ctionality

### DIFF
--- a/include/votca/tools/random2.h
+++ b/include/votca/tools/random2.h
@@ -23,8 +23,8 @@
      uniformly distributed in the range [0,1)
 *************************************************/
 
-#ifndef _RANMARS2_H_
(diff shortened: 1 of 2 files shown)


JensWehner:
yeah but it is not used at the moment and if it would be used the state would probably stored with the rest of the simulation. Also the form is very C like, so I would implement it if needed.

include/votca/tools/random2.h
```diff
@@ -26,12 +26,8 @@
 #ifndef _RANMARS2_H_
 #define _RANMARS2_H_
```
junghans: Maybe we should rename the guards while we are at rewritting.

PR #52, opened 2018-07-27
Enhancement add bounds checking matrix

### LINKED ISSUE #51: Unsafe public facing matrix api
The setters and getters in the matrix class do not check the scope of the indices against the size of the matrix. This is dangerous as a wrong index could easily be missed and could be potentially hard to debug as it might lead to random errors.

JoshuaSBrown:
Check the first commit for the actual code that I changed, Everything in the second commit is simply a result of google formatting changes. 

Fix #51

### DIFF
--- a/include/votca/tools/matrix.h
+++ b/include/votca/tools/matrix.h
@@ -253,7 +253,7 @@ class matrix {
 };
 
 inline void matrix::checkBounds_(const byte_t row_or_col) const {
-  assert( row_or_col > 2 || row_or_col < 0);
+  assert( row_or_col > -1 && row_or_col < 3);
 } 
 
 inline matrix &matrix::operator=(const double &v) {
--- a/include/votca/tools/matrix.h
+++ b/include/votca/tools/matrix.h
@@ -18,7 +18,6 @@
 #ifndef _VOTCA_TOOLS_MAT_H
 #define _VOTCA_TOOLS_MAT_H
 
-#include <boost/lexical_cast.hpp>
 #include <cassert>
 #include "floatingpointcomparison.h"
 #include "ostream"
@@ -256,7 +255,7 @@ class matrix {
 };
 
 inline void matrix::checkBounds_(const byte_t row_or_col) const {
-  assert( row_or_col < boost::lexical_cast<byte_t>(3));
+  assert( row_or_col < static_cast<byte_t>(3));
 } 
 
 inline matrix &matrix::operator=(const double &v) {
--- a/include/votca/tools/matrix.h
+++ b/include/votca/tools/matrix.h
@@ -256,7 +256,7 @@ class matrix {
 };
 
 inline void matrix::checkBounds_(const byte_t row_or_col) const {
-  assert( row_or_col > boost::lexical_cast<byte_t>(-1) && row_or_col < boost::lexical_cast<byte_t>(3));
+  assert( row_or_col < boost::lexical_cast<byte_t>(3));
 } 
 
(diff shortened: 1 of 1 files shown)


JoshuaSBrown:
This is a point worth discussing though. If performance is an issue we could replace all bounds checking with a MACRO that by default does nothing unless you pass in a flag to turn it on? This would not just increase performance in these two classes e.g. 

    getElem(int ind){
         return data_[ind];
    }

vs something like this

    getElem(int ind){
        PERFORMANCE_BOUNDS_TEST(ind<0 || ind>2, "calling index out of bounds)
        return data_[ind];
     }
Is this the only place in the code other than vec where this is an issue? 

The problem I have is that it is not immediately clear that the code was implemented this way for performance issues.

JensWehner:
I do not like the macros for many reasons.  You can use asserts if you want because they are disabled in Release mode. I do not know about any other linear algebra library that does range checking in Release mode. My point is also more why invest time in vec and matrix if we will replace it in the not too far future with something which offers the functionality

JoshuaSBrown:
If it is an issue of just matrix and vec classes then there is not much point, but as a discussion about performance it is relevant. The only issue I have with assertions is that it is also not always obvious why they are used in place of an if statement. I'm not a fan of macros either but they provide a way to spell out the reason they are there.

JoshuaSBrown:
Oh, and on second thought we don't need to use a macro we could simply use a function.

include/votca/tools/matrix.h
```diff
@@ -33,283 +33,303 @@ namespace tools {
...
 
-  /* provided the matrix a diagonalizes it and returns the eigenvalues
-     lambda0 = a[0][0], lambda1 = a[1][1], lambda2= a[2][2], ...
-     as well as the corresponding eigenvectors v[][0], v[][1], v[][2]
-   */
-  int cjcbi(matrix &a, matrix &v, double eps = 1e-10, int jt = 100);
+ private:
+  double _m[9];
+  void checkBounds_(const byte_t row_or_col) const {
+    if (row_or_col > 2 || row_or_col < 0) {
```
junghans: Yeah, but make this an `assert()`.

junghans:
@JoshuaSBrown, I can reproduced this:
```
cmake -DENABLE_TESTING=ON -DCMAKE_BUILD_TYPE=Debug ..
make
ctest -R Run_csg_fmatch -V
...
bad lexical cast: source type value could not be interpreted as target
```

include/votca/tools/matrix.h
```diff
@@ -33,283 +36,301 @@ namespace tools {
...
-     as well as the corresponding eigenvectors v[][0], v[][1], v[][2]
-   */
-  int cjcbi(matrix &a, matrix &v, double eps = 1e-10, int jt = 100);
+ private:
+  double _m[9];
+  void checkBounds_(const byte_t row_or_col) const;
+};
+
+inline void matrix::checkBounds_(const byte_t row_or_col) const {
+  assert( row_or_col > boost::lexical_cast<byte_t>(-1) && row_or_col < boost::lexical_cast<byte_t>(3));
```
JensWehner: why byte? Additionally the assert should have a message E.g. "Access beyond array boundary"

ruehle: Also why lexical_cast, and why -1 with byte which is unsigned

JoshuaSBrown: I guess there is no need to check the lower bound than. Thanks for pointing that out.

PR #61, opened 2018-08-12
Fixed memory error

### LINKED ISSUE #60: unit_test_graphalgorithm fails witn gcc-4.8.5 and boost-1.58.0
```
10/21 Test #10: unit_test_graphalgorithm ............***Failed    0.01 sec
Running 3 test cases...
unknown location(0): fatal error in "decoupleIsolatedSubGraphs_algorithm_test": memory access violation at address: 0x00000010: no mapping at fault address
/home/travis/votca/src/tools/src/tests/test_graphalgorithm.cc(133): last checkpoint

*** 1 failure detected in test suite "graphalgorithm_test"
```
details see here: https://api.travis-ci.org/v3/job/414739327/log.txt

JoshuaSBrown:
Error was caused from using a for range loop with a vector while also skipping vertices occasionally. The for loop was not ending correctly and thus was trying to access memory outside of the vector memory block. 

Fix #60

### DIFF
--- a/src/libtools/graph.cc
+++ b/src/libtools/graph.cc
@@ -18,6 +18,7 @@
  */
 
 #include <algorithm>
+#include <cassert>
 #include <string>
 #include <votca/tools/graph.h>
 
@@ -106,7 +107,10 @@ void Graph::setNode(std::pair<int, GraphNode> p_gn) {
   setNode(p_gn.first, p_gn.second);
 }
 
-GraphNode Graph::getNode(int vert) { return nodes_[vert]; }
+GraphNode Graph::getNode(int vert) { 
+  assert(nodes_.count(vert));
+  return nodes_[vert]; 
+}
 
 vector<pair<int, GraphNode>> Graph::getNodes(void) {
   vector<pair<int, GraphNode>> vec_nodes;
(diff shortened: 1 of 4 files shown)


src/libtools/graph.cc
```diff
@@ -106,7 +107,10 @@ void Graph::setNode(std::pair<int, GraphNode> p_gn) {
   setNode(p_gn.first, p_gn.second);
 }
 
-GraphNode Graph::getNode(int vert) { return nodes_[vert]; }
+GraphNode Graph::getNode(int vert) { 
+  assert(nodes_.count(vert));
```
jkrajniak: Keep in mind that `assert` works only when `NDEBUG` macro is not defined.

JoshuaSBrown:
Addressing issue: https://github.com/votca/tools/issues/60

src/tests/test_graphalgorithm.cc
```diff
@@ -71,7 +72,7 @@ BOOST_AUTO_TEST_CASE(single_network_algorithm_test) {
     gb_v2.setStartingVertex(2);
     BOOST_CHECK_THROW(singleNetwork(g, gb_v2), invalid_argument);
   }
-
+/*
```
jkrajniak: why commented?

src/libtools/graph.cc
JoshuaSBrown: 👍

src/tests/test_graphalgorithm.cc
JoshuaSBrown: Forgot to uncomment.

PR #66, opened 2018-08-23
property now outputs empty xml property strings without offset

JensWehner:
before: `<value attr1="bla">              </value>`
now:    `<value attr1="bla"/>`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ endif(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/.git)
 find_package(Threads REQUIRED)
 set(THREAD_LIBRARIES ${CMAKE_THREAD_LIBS_INIT})
 
-find_package(Boost 1.39.0 REQUIRED COMPONENTS program_options filesystem system )
+find_package(Boost 1.57.0 REQUIRED COMPONENTS program_options filesystem system )
 include_directories(${Boost_INCLUDE_DIRS})
 set (BOOST_CFLAGS_PKG "-I${Boost_INCLUDE_DIRS}")
 set(BOOST_LIBS_PKG "-L${Boost_LIBRARY_DIRS}")
--- a/netbeans/libtools/nbproject/configurations.xml
+++ b/netbeans/libtools/nbproject/configurations.xml
@@ -29,6 +29,7 @@
       <itemPath>../../include/votca/tools/objectfactory.h</itemPath>
       <itemPath>../../include/votca/tools/parsexml.h</itemPath>
       <itemPath>../../include/votca/tools/property.h</itemPath>
+      <itemPath>../../include/votca/tools/random2.h</itemPath>
       <itemPath>../../include/votca/tools/rangeparser.h</itemPath>
       <itemPath>../../include/votca/tools/spline.h</itemPath>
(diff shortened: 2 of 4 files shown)


ruehle:
why not
<value attr1="bla"/>

CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ endif(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/.git)
 find_package(Threads REQUIRED)
 set(THREAD_LIBRARIES ${CMAKE_THREAD_LIBS_INIT})
 
-find_package(Boost 1.39.0 REQUIRED COMPONENTS program_options filesystem system )
+find_package(Boost 1.57.0 REQUIRED COMPONENTS program_options filesystem system )
```
JoshuaSBrown: Does this mean that Boost 1.57.0 is the minimum required version?

src/libtools/property.cc
```diff
@@ -179,88 +177,87 @@ void PrintNodeTXT(std::ostream &out, Property &p, const int start_level, int lev
...
+      string cAttributeValue = color->Green();
+      string cReset = color->Reset();
+
+      // print starting only from the start_level (the first node (level 0) can be <> </>)
+      if (level >= start_level) {
+        // print the node name
+        out << indent << offset << "<" << cKey << p.name() << cReset;
+        // print the node attributes 
+        for (ia = p.firstAttribute(); ia != p.lastAttribute(); ++ia) {
+          out << " "
```
JoshuaSBrown: Can you clean up the indents?

src/libtools/property.cc
```diff
@@ -179,88 +177,87 @@ void PrintNodeTXT(std::ostream &out, Property &p, const int start_level, int lev
...
+      }
+
+      // continue iteratively through the rest of the nodes
+      for (Property& prop : p) {
+        level++;
+        if (level > start_level) offset += "\t";
+        PrintNodeXML(out, prop, piom, level, offset);
+        if (level > start_level) offset.resize(offset.size() - 1);
+        level--;
+      }
```
JoshuaSBrown: This doesn't really matter but just for your info: I believe preincremeinting is supposed to be faster than post though some people argue it doesn't really matter. 
https://softwareengineering.stackexchange.com/questions/358725/pre-increment-vs-post-increment

```
--level;
++level;
```

netbeans/libtools/nbproject/configurations.xml
```diff
@@ -40,7 +39,6 @@
       <itemPath>../../include/votca/tools/vec.h</itemPath>
       <itemPath>../../include/votca/tools/version.h</itemPath>
       <itemPath>../../include/votca/tools/votca_config.h.in</itemPath>
-      <itemPath>../../include/votca/tools/votca_gsl_boost_ublas_matrix_prod.h</itemPath>
```
JoshuaSBrown: We are no longer using ublas because we switched to eigen right?

JensWehner: yes

src/libtools/property.cc
JensWehner: for simple data types it makes no difference

CMakeLists.txt
JensWehner: yes. the rest of votca has that already

PR #68, opened 2018-08-27
removed couts

JensWehner:


junghans:
Please don't use `JensWehner-patch-1` branch name!

PR #72, opened 2018-09-26
property can now convert to some eigen structures

JensWehner:
-Property can now directly convert to Eigen::Vector3d and Eigen::VectorXd

I tried to get a more general implementation but as far as I know I cannot partially specialize the templated function


junghans:
And how about a test?

JensWehner:
Good idea, there is actually a big problem with this commit.

JensWehner:
Or at least I think so. We have big problem compiling with mkl now, because CTP uses cblas from GSL and then both versions get confused, with each other.

JensWehner:
the intel blas and the gsl blas

PR #74, opened 2018-09-26
changed order so parser does not get initialized if file does not exist

JensWehner:
avoids memory leak, because otherwise if file does not exist parser is not freed.


JensWehner:
the edge fix is just for -Wpedantic warnings

JoshuaSBrown:
Well this might also explain the seg fault I was getting with csg_stat when testing with molecule-refactor branch in the csg repo.

PR #77, opened 2018-10-05
Runtime error identity class

JoshuaSBrown:


JoshuaSBrown:
The correct include was missing from the identity class.

PR #78, opened 2018-10-05
Added correct include for name class

JoshuaSBrown:


JoshuaSBrown:
runtime_error requires the <stdexcept> include

PR #82, opened 2018-10-24
FindEIGEN3.cmake: update from upstream

junghans:
Related to votca/votca#92


JensWehner:
that is a much better solution than mine (mine did not fully work). Thanks

PR #83, opened 2018-12-21
Added more test to datacollection

JoshuaSBrown:
Tested methods of data collection class more thoroughly.


JoshuaSBrown:
@junghans did you look over this?

JoshuaSBrown:
It is a very general class used to store arrays of data. I think that Victor made this class. Though I think you could do similar things with a map of vectors. It has some other functionality such as selecting columns of data using wild card comparison of the columns name.

PR #89, opened 2019-01-04
Set a default value for the number of bins to 100, it was being used …

JoshuaSBrown:
…without being initialized

### DIFF
--- a/src/libtools/histogramnew.cc
+++ b/src/libtools/histogramnew.cc
@@ -26,6 +26,7 @@ namespace votca {
       _min = _max = _step = _step_p = 0;
       _weight = 1.;
       _periodic = false;
+      _nbins = 100;
     }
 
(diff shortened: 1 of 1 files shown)


JoshuaSBrown:
Your saying the constructor takes a parameter that you must provide to initialize it?

JensWehner:
no , 
you can write 

`int _size=0;`

in the private section and than you have to do nothing in the constructor

JoshuaSBrown:
Ok, so moving initialization to the header file in the private section and setting it to a default of 0 instead of 100?

JensWehner:
I am working on something like it atm.

JensWehner:
100 is a good number

JoshuaSBrown:
Ok, I've assigned you to the merge, so I don't do anything else.

PR #90, opened 2019-01-04
Hist refactor

JensWehner:
- Deleted all copy constructors for histogram and table, so data is genuinely copied
- Histogram now only support either periodic or normal, not both at the same time
- Removed `_weight` variable because it is always 1

### DIFF
--- a/include/votca/tools/histogramnew.h
+++ b/include/votca/tools/histogramnew.h
@@ -166,7 +166,6 @@ class HistogramNew {
   double _min=0;
   double _max=0;
   double _step=0;
-  double _step_p=0;
   bool _periodic=false;
   int _nbins=100;
   Table _data;


include/votca/tools/histogramnew.h
```diff
@@ -144,7 +137,7 @@ class HistogramNew {
    * \brief get the grid of histogram
    * \return step per bin
    */
-  double getStep() const { return _periodic ? _step_p : _step; }
+  double getStep() const { return _step; }
```
JoshuaSBrown: Does this not cause problems with the unit tests? I would also suggest you check the csg tests as well as some of them might depend on this functionality.

JensWehner: I ran all tests locally

include/votca/tools/histogramnew.h
```diff
@@ -169,19 +162,14 @@ class HistogramNew {
   void setPeriodic(bool periodic) { _periodic = periodic; }
```
JoshuaSBrown: I would recommend throwing an error if the periodicity is changed at some point after data is put in the table. So that some one is not accessing data that they think is periodic when it is no longer.

JensWehner: or we reset the data and initialize again.

JoshuaSBrown: Reset when a setting is changed?

JensWehner: a) throwing an error, requires knowledge if processing has happened already
b) requires that we initialize before setting periodicity
c) the current version works with the current code (I hope)

JoshuaSBrown: It will, probably, but I think it might be inviting future problems, and that would be a hard bug to find. I think some kind of check should be in place. 

If you have to manually reinitialize after changing a setting that is problematic, unless the code fails catastropically you might very well forget to reinitialize.

JensWehner: Can we make the periodic bool part of the initialisation, so it would all be together, maybe as a default=false, I would also rename initliasation to SetupBins

JoshuaSBrown: Sounds good to me.

JensWehner: Unfortunately, Changing the Interface of histogram is problematic as it involves csg and xtp and tools. Can we merge it as it is and then I'll change it later? @junghans @JoshuaSBrown

JoshuaSBrown:
Ok

PR #91, opened 2019-01-04
[WIP] Clean namespace from using

JoshuaSBrown:


JensWehner:
I think we should merge into stable not master, atm. @junghans

JoshuaSBrown:
@junghans ?

junghans:
Is that PR something we need right now to fix the memory leaks? Or can we wait with merging this?

JoshuaSBrown:
No this is not a memory leak problem. It is technical debt, which is why I picked master.

junghans:
Ok, it seems to breaks the code in master of ctp.

JensWehner:
@JoshuaSBrown @junghans This can be closed, as all using namespace were removed already.

junghans:
Ok

PR #93, opened 2019-01-12
Refactor graph class add reduced graph functionality

JoshuaSBrown:
The purpose of this branch is to introduce new functionality that reduces the complexity of a graph by removing any vertex that has a degree of 2. By exploring a reduced graph instead of a full graph insight can still be gained into the topology of the graph but with better performance. The edges of the reduced graph can be expanded if needed to determine how they correspond to the full graph. Take:

<pre>
1 - 2 - 3 - 4 - 5 - 6
    |       |
    7 - 8 - 9 
</pre>
This would be reduced to 
 
<pre>
1 - 2 - 4 - 6
    | _ | 
</pre>
A total of 4 vertices with 4 edges as opposed to 9 vertices and 9 edges.


junghans:
Can we delete that branch?

PR #95, opened 2019-01-13
Feature explore branch algorithm

JoshuaSBrown:
This branch includes an algorithm that will explore a chosen branch within a graph. If you were to choose a vertex within the graph and pick one of the edges connected to vertex and explore all the way down the tree from that point on you would explore the branch. In the case that the exploration returns back to the vertex it will not continue past the starting point. 

E.g. If I have a graph
<pre>
1 - 2 - 3 - 4 
|   |
5 - 6
</pre>

And my starting vertex is two and I set the branch to explore by specifying edge 1 - 2. I would end up with a branch that looked like this:

<pre>
1 - 2 
|   |
5 - 6
</pre>

The purpose of doing this is it provides the basis for defining functionality to resolve this structural components of molecules. 

Should only be merged after: https://github.com/votca/tools/pull/94

### DIFF
--- a/include/votca/tools/edge.h
+++ b/include/votca/tools/edge.h
@@ -59,7 +59,7 @@ class Edge {
    * If both ends of the edge point to the same vertex than it is considered a
    * loop.
    **/
-  bool loop() { return vertices_.front() == vertices_.back(); }
+  bool loop() const { return vertices_.front() == vertices_.back(); }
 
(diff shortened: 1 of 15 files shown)


JoshuaSBrown:
@JensWehner Nothing to add for this pull request?

JensWehner:
I am not that fast.

PR #102, opened 2019-01-30
improved warnings for elements

JensWehner:
Throws error if you request mass of an unknown element

### DIFF
--- a/include/votca/tools/elements.h
+++ b/include/votca/tools/elements.h
@@ -42,13 +42,18 @@ namespace votca {
         class Elements {
         public:
 
+            Elements(): _filled_VdWChelpG(false),_filled_VdWMK(false),_filled_NucCrg(false),
+                        _filled_CovRad(false),_filled_Mass(false),_filled_EleNum(false),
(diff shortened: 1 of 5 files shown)


include/votca/tools/elements.h
```diff
@@ -42,48 +42,53 @@ namespace votca {
...
             /// be found here [Singh, U. Chandra (04/01/1984). "An approach to computing
             /// electrostatic charges for molecules". Journal of computational chemistry
             /// 5 (2), 129, 1984]. The VdWMK method will return the radii used to define
             /// where the electrostatic grid starts.
-            const double &getVdWMK(std::string name);
+            double getVdWMK(std::string name);
 
             /// Return the Nuclear charges of each atom. H - 1, He - 2, Na - 3 etc...
-            const double &getNucCrg(std::string name);
+            double getNucCrg(std::string name);
```
JoshuaSBrown: Do we want to go ahead and expand these function names, getNuclearCharge, getElementNumber, or do you want to just wait?

JensWehner: We can do that but not in this PR.. for it we have to make a new branch in every repo

PR #103, opened 2019-01-30
Bugfix findsubgraphs

JensWehner:
rewrote subdivison and it fixes https://github.com/votca/csg/issues/319

do not know exactly why

### DIFF
--- a/src/libtools/graphalgorithm.cc
+++ b/src/libtools/graphalgorithm.cc
@@ -33,17 +33,6 @@ namespace tools {
  * Internal Functions *
  **********************/
 
-list<int> vectorToList_(vector<int> values) {
-  list<int> values_list;
-  copy(values.begin(), values.end(), back_inserter(values_list));
-  return values_list;
-}
-
-vector<Edge> edgeSetToVector_(set<Edge> edges) {
-  vector<Edge> edges_vector;
-  copy(edges.begin(), edges.end(), back_inserter(edges_vector));
-  return edges_vector;
-}
 
 /********************
  * Public Functions *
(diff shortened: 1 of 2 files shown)


src/libtools/graphalgorithm.cc
```diff
@@ -248,52 +237,43 @@ ReducedGraph reduceGraph(Graph graph) {
...
+ //bool vector to see if vertex is already part of graph
+ std::vector<bool> vertex_analysed=std::vector<bool>(vertices.size(),false);
 
-  list<int>::iterator vertices_iterator = vertices_list.begin();
-  while (vertices_iterator != vertices_list.end()) {
-    unordered_map<int, GraphNode> sub_graph_nodes;
+ std::vector<Graph> subGraphs;
+ unsigned i=0;
+ while(i<vertices.size()){
+    if(vertex_analysed[i]){i++;continue;}
```
JoshuaSBrown: This if statement should probably be a while, or something of the sort. You do not want to explore any of the vertices if they have already been explored. Stop when when you hit a vertex that has not been explored or there are none left to explore.

JoshuaSBrown: while(vertex_analysed[i] && i<vertices.size()){ i++;continue}

Or something like it, then you will have to check that you have not reached the end of the available vertices before you continue. 

Nevermind, I didn't realize what the continue keyword did.

JoshuaSBrown: Ya, I think in the future I will avoid deleting elements in a loop while iterating over them.

PR #108, opened 2019-02-07
switched lexical_cast<double> to stod

junghans:
Fix #107


junghans:
@JoshuaSBrown please review!

PR #111, opened 2019-02-12
Added method for checking if a string matches an elements symbol or name

JoshuaSBrown:
Adds a simple method to check if a string is the same as an elements name or its symbol.


JensWehner:
can we change clang format to only format the code that was changed not the whole file? It is really hard to do code review.

JoshuaSBrown:
I would not know how to do that. I can go ahead and format the hole repo. Then this should not be as difficult.

src/libtools/elements.cc
```diff
@@ -148,7 +148,7 @@ bool Elements::isEleFull(std::string fullname) {
 }
 
 bool Elements::isEleShort(std::string shortname) {
-  if (!this - _filled_EleShort) {
+  if (!this->_filled_EleShort) {
```
JensWehner: I still do not get it, you should fill FillEleShort not long, this is confusing. I do not understand

JensWehner: I get the part with the function name, but the bools should not disagree

JoshuaSBrown: Oh, right this one is backwards.

src/libtools/elements.cc
```diff
@@ -25,732 +26,767 @@ using namespace std;
...
-    }
-    try{
-        return _NucCrg.at(name);
-    }catch(const std::out_of_range& oor){
-        throw std::runtime_error("Nuclearcharge of element "+name+" not found.");
-    }
+bool Elements::isElement(std::string name) {
+  if (!this->_filled_EleShort) {
+    this->FillEleShort();
+    _filled_EleShort;
```
JensWehner: must be set to true

src/libtools/elements.cc
```diff
@@ -25,732 +26,767 @@ using namespace std;
...
+  }
+  if (!this->_filled_EleFull) {
+    this->FillEleFull();
+    _filled_EleFull;
```
JensWehner: dito

JoshuaSBrown:
@JensWehner Hey can you approve this?

src/libtools/elements.cc
```diff
@@ -25,6 +26,19 @@ using namespace std;
 /*************************
  * Public Facing Methods *
  *************************/
+bool Elements::isElement(std::string name) {
+  if (!this->_filled_EleShort) {
+    this->FillEleShort();
+    _filled_EleShort = true;
+  }
+  if (!this->_filled_EleFull) {
```
JensWehner: call the other to methods below here instead of writing it again

JoshuaSBrown:
@JensWehner I think I have now addressed your concerns.

PR #112, opened 2019-02-12
Format all files

JoshuaSBrown:
Formatted all the files with clang-formatter.


JoshuaSBrown:
@JensWehner Here I did not make any changes but formatting. Any merges from this point on should clearly indicate which files have been changed.

PR #116, opened 2019-02-16
Improve codecov coverage

junghans:
Part of https://github.com/votca/votca/pull/132


junghans:
The logfile are too long, but I added a fix for that in votca/votca#132.

junghans:
This doesn't report status yet, we need to merge https://github.com/votca/votca/pull/135 first.

PR #120, opened 2019-02-26
votca_compare: move to python3

junghans:
Part of https://github.com/votca/csg/issues/341


JensWehner:
We should remove the "import print from future" stuff to

PR #127, opened 2019-03-22
Add memory testing

JoshuaSBrown:
Uses valgrind if it is found to test the unit tests to ensure there are no intrinsic memory errors. @junghans how do we ensure that valgrind is installed on the gitlab runners and travis I did not see where you were configuring cmake and other dependencies in the .ci-.. files?


CMakeLists.txt
```diff
@@ -42,7 +42,12 @@ include(GNUInstallDirs)
 
 option(ENABLE_TESTING "Build and enable testing stuff" OFF)
 if(ENABLE_TESTING)
+  include(CTest)
```
junghans: Why do we need this?

CMakeLists.txt
```diff
@@ -42,7 +42,12 @@ include(GNUInstallDirs)
...
   enable_testing()
+  find_program(VALGRIND "valgrind")
+  if(NOT VALGRIND)
```
junghans:
```suggestion
  find_program(VALGRIND_EXECUTABLE "valgrind")
  find_package_handle_standard_args(VALGRIND REQUIRED_VARS VALGRIND_EXECUTABLE)
```
CMake default message is better!

src/tests/CMakeLists.txt
```diff
@@ -38,4 +38,7 @@ foreach(PROG
   # run tests for tools (for coverage) as well
   set_tests_properties(unit_${PROG} PROPERTIES LABELS "tools;votca")
 
+  if(VALGRIND)
+    add_test(NAME memory_${PROG} COMMAND "${VALGRIND}" --tool=memcheck --leak-check=yes --show-reachable=yes --num-callers=20 --track-fds=yes --error-exitcode=1 $<TARGET_FILE:unit_${PROG}>)
```
junghans: Please label the new test as well.
```suggestion
    add_test(NAME memory_${PROG} COMMAND "${VALGRIND}" --tool=memcheck --leak-check=yes --show-reachable=yes --num-callers=20 --track-fds=yes --error-exitcode=1 $<TARGET_FILE:unit_${PROG}>)
    set_tests_properties(memory_${PROG} PROPERTIES LABELS "tools;votca")
```

CMakeLists.txt
JoshuaSBrown: Yes, but that command is only supported in the most recent versions of cmake, I have cmake 3.10 and it is not recognized.

CMakeLists.txt
JoshuaSBrown: I guess not.

CMakeLists.txt
JoshuaSBrown: Never mind, I was missing an include.

src/tests/CMakeLists.txt
```diff
@@ -38,4 +38,8 @@ foreach(PROG
   # run tests for tools (for coverage) as well
   set_tests_properties(unit_${PROG} PROPERTIES LABELS "tools;votca")
 
+  if(VALGRIND_EXECUTABLE)
```
junghans:
```suggestion
  if(VALGRIND_FOUND)
```
`find_package_handle` defines that.

PR #130, opened 2019-03-26
Attempt at fixing compile problem

JoshuaSBrown:
Attempt to fix problem:

https://groups.google.com/forum/#!msg/votca/LGjDJLwuFgI/EnM4zFvJBQAJ

### DIFF
--- a/include/votca/tools/graph.h
+++ b/include/votca/tools/graph.h
@@ -124,8 +124,8 @@ class Graph : public EdgeContainer {
  *
  * sort(vec_pr_gn.begin(),vec_pr_gn.end(),cmpVertNodePair);
  */
-bool cmpVertNodePair(const std::pair<int, GraphNode> gn1_pr,
-                     const std::pair<int, GraphNode> gn2_pr);
+bool cmpVertNodePair(const std::pair<int, GraphNode>& gn1_pr,
+                     const std::pair<int, GraphNode>& gn2_pr);
 }  // namespace tools
 }  // namespace votca
 #endif  // _VOTCA_TOOLS_GRAPH_H
--- a/include/votca/tools/graph.h
+++ b/include/votca/tools/graph.h
@@ -1,5 +1,5 @@
 /*
- *            Copyright 2009-2018 The VOTCA Development Team
+ *            Copyright 2009-2019 The VOTCA Development Team
  *                       (http://www.votca.org)
  *
  *      Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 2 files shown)


JoshuaSBrown:
It looks like the pull request is building with master instead of stable...

include/votca/tools/graph.h
```diff
@@ -124,8 +124,8 @@ class Graph : public EdgeContainer {
  *
  * sort(vec_pr_gn.begin(),vec_pr_gn.end(),cmpVertNodePair);
  */
-bool cmpVertNodePair(std::pair<int, GraphNode> gn1_pr,
-                     std::pair<int, GraphNode> gn2_pr);
-}
-}
+bool cmpVertNodePair(const std::pair<int, GraphNode> gn1_pr,
+                     const std::pair<int, GraphNode> gn2_pr);
```
JensWehner: should be references not copies

src/libtools/graph.cc
```diff
@@ -83,18 +83,18 @@ vector<int> Graph::getVerticesMissingNodes(void) {
...
+  for (auto neigh_vert : neigh_vertices) {
+    auto node_pr = pair<int, GraphNode>(neigh_vert, nodes_[neigh_vert]);
     neigh_vertices_pr.push_back(node_pr);
   }
   return neigh_vertices_pr;
 }
 
-void Graph::setNode(int vert, GraphNode gn){
-  if(nodes_.count(vert)){
+void Graph::setNode(int vert, GraphNode gn) {
```
JensWehner: should this be a copy, or if graph is not a pdo type this should be a const reference

src/libtools/graph.cc
```diff
@@ -140,9 +140,10 @@ ostream& operator<<(ostream& os, const Graph g) {
   return os;
 }
 
-bool cmpVertNodePair(pair<int, GraphNode> gn1_pr, pair<int, GraphNode> gn2_pr) {
+bool cmpVertNodePair(const pair<int, GraphNode> gn1_pr,
```
JensWehner: this should be a const reference

src/libtools/graph.cc
JoshuaSBrown: what is 'pdo' ?

JoshuaSBrown: I'm not fixing this in stable, because it might start me down a rabbit hole and it is not what is causing the error.

JoshuaSBrown:
gitlab still building against master...

junghans:
The branch name needs to be `for/stable/...`, but don't worry I just push the commit on another branch, so that should trigger the build without the need for a new PR.

JoshuaSBrown:
oh right, I forgot. My bad.

PR #135, opened 2019-03-30
[WIP] Electrostatics joshs development

JoshuaSBrown:
1. Added pragma once (compiler performance)
2. Added UnitConverter (standardize unit conversion technique)
3. Added StructureParameter class ( used for standarizing construction of arbitrary classes with a single argument, thus making templates easier to create)


include/votca/tools/akimaspline.h
```diff
@@ -15,6 +15,7 @@
  *
  */
 
+#pragma once
```
JensWehner: why add this?

include/votca/tools/application.h
```diff
@@ -15,6 +15,7 @@
  *
  */
 
+#pragma once
```
JensWehner: why?

JensWehner: I see the advantages, but I want a discussion before we start doing this.

include/votca/tools/structureparameters.h
```diff
@@ -0,0 +1,139 @@
...
+
+#include "unitconverter.h"
+#include <Eigen/Dense>
+#include <boost/any.hpp>
+#include <cassert>
+#include <unordered_map>
+
+namespace votca {
+namespace tools {
+
```
JensWehner: Can you explain this a bit more?

include/votca/tools/structureparameters.h
```diff
@@ -0,0 +1,139 @@
...
+    assert(parameters_.count(parameter) &&
+           "StructureParameter is not stored in StructureParameters class");
+    assert(typeid(T) == parameters_.at(parameter).type() &&
+           "Cannot return boost any value from parameters class because it is "
+           "not being cast to the correct type");
+    return boost::any_cast<T>(parameters_.at(parameter));
+  }
+
+ private:
+  void convertParameterIfPossible_(const StructureParameter parameter);
```
JensWehner: why no return type, where does it convert?

JensWehner: why not a const reference?

include/votca/tools/unitconverter.h
```diff
@@ -0,0 +1,223 @@
...
+      case DistanceUnit::meters:
+        return 1E10;
+      case DistanceUnit::centimeters:
+        return 1E8;
+      case DistanceUnit::nanometers:
+        return 10.0;
+      case DistanceUnit::angstroms:
+        return 1.0;
+      case DistanceUnit::bohr:
+        return 0.529177;
```
JensWehner: I am not sure if this is enough digits

include/votca/tools/akimaspline.h
JoshuaSBrown: "In the C and C++ programming languages, #pragma once is a non-standard but widely supported preprocessor directive designed to cause the current source file to be included only once in a single compilation. Thus, #pragma once serves the same purpose as include guards, but with several advantages, including: less code, avoidance of name clashes, and sometimes improvement in compilation speed."

include/votca/tools/structureparameters.h
JoshuaSBrown: I am still unsure if I wan to keep this method. I thought it could be used to automatically convert a CSG parameters to XTP parameters if one is missing. For instance if I have the CSG_Position of an object but not the XTP_Position (Units are different), I could automatically convert the CSG_Position to the XTP_Position. The problem with this is that the class then needs to know about what units are used in xtp vs csg, what do you think?

include/votca/tools/unitconverter.h
JoshuaSBrown: Ok will see if I can improve all the units, higher sf etc.

include/votca/tools/structureparameters.h
JoshuaSBrown: So the comments I thought were quite explicit, if you can target which part you are a little confused about I can try to clear it up though.

JensWehner: why structure parameters and not simply units?

JensWehner: ignore that please. I still do not get the advantage of the structure parameter class

JoshuaSBrown: Units is too general, an id is not really a unit any way (int would be the unit), these are meant to be parameters passed to a class that is related to the structure, molecules, fragments, motifs etc.

include/votca/tools/structureparameters.h
JoshuaSBrown: This method has been removed.

include/votca/tools/unitconverter.h
JoshuaSBrown: This has been fixed.

include/votca/tools/unitconverter.h
```diff
@@ -30,14 +30,38 @@ enum MassUnit {
...
+enum ChargeUnit { e, coulombs };
+
+enum VelocityUnit {
+  angstroms_per_femtosecond,
+  angstroms_per_picosecond,
+  nanometers_per_picosecond
+};
 
+enum ForceUnit {
+  kilocalories_per_mole_ansgtrom,
```
JensWehner: spelling

JensWehner: why not hartree/bohr?

include/votca/tools/unitconverter.h
```diff
@@ -82,36 +106,90 @@ class UnitConverter {
...
+        return 1.0;
     }
     return 0.0;
   }
 
   /// All energies in terms of electron volts
   constexpr double getEnergyValue_(const EnergyUnit& enum_type) const noexcept {
     switch (enum_type) {
+      case EnergyUnit::kilojoules_per_mole:
+        return 96.485;
```
JensWehner: that is not a lot of digits

src/libtools/spline.cc
```diff
@@ -34,7 +34,9 @@ int Spline::GenerateGrid(double min, double max, double h) {
   }
   _r[i] = max;
   _f.resize(_r.size());
```
JensWehner: is there a unit test for this

JensWehner: because it changes behaviour, I am not saying it is wrong, it is probably better this way.

include/votca/tools/unitconverter.h
JoshuaSBrown: Good catch

include/votca/tools/unitconverter.h
JoshuaSBrown: good catch

src/libtools/spline.cc
JoshuaSBrown: I will probably have to create one, I know I have made a lot of changes that may not be thoughtfully tested, I wandering when this first gets released it should probably be released as an alpha, at least until we can get our code coverage up.

include/votca/tools/unitconverter.h
JoshuaSBrown: What's wrong with the spelling?

JensWehner: ansgstrom, angstrom

PR #142, opened 2019-04-30
Added parameters class

JoshuaSBrown:
Description has been added to the parameters.h class file. Essentially establishes a standardized method for initializing class constructors.


JensWehner:
can you describe a bit more what the PR is about?

JoshuaSBrown:
@JensWehner Take a look now.

JensWehner:
For me the name is confusing, a bit to generic, could we call it particle_parameter or so, because I think it would be a bad idea if someone uses it outside of file readers and writers

JensWehner:
Otherwise I like it

JoshuaSBrown:
How about topology Parameter because it can be used with segments molecules etc...?

JensWehner:
structure_parameter? otherwise topology_parameter is fine

PR #143, opened 2019-04-30
Update compare function for floats

JensWehner:


junghans:
```
The following tests FAILED:
	 41 - integration_Compare_csg_reupdate_output (Failed)
	 43 - integration_Compare_csg_fmatch_output (Failed)
	 45 - integration_Compare_csg_fmatch_3body_output1 (Failed)
	 46 - integration_Compare_csg_fmatch_3body_output2 (Failed)
	 47 - integration_Compare_csg_fmatch_3body_output3 (Failed)
	 49 - integration_Compare_csg_map_output (Failed)
	 51 - integration_Compare_csg_stat_output1 (Failed)
	 52 - integration_Compare_csg_stat_output2 (Failed)
	 54 - integration_Compare_csg_stat_angular_output (Failed)
	 56 - integration_Compare_csg_stat-imc_output (Failed)
	 57 - integration_Compare_csg_stat-imc_output_2 (Failed)
	 59 - integration_Compare_csg_resample_akima_output (Failed)
	 61 - integration_Compare_csg_resample_cubic_output (Failed)
	 63 - integration_Compare_csg_resample_linear_output (Failed)
	 65 - integration_Compare_csg_resample_cubicfit_output (Failed)
```

JensWehner:
We have two problems:

a) the new comparison is stricter than the old one

b) we also compare the errors to a ref error of 0. 

example: 
ref_file:
#value # error
1300 0
50 0
2 0

result_file:
#value # error
1300 1e-3
50 1e-5
2 1e-6

the relative errors for the results are compareable but the absolute error for 1300 is too big because it is larger than the threshold 

The first problem is easily solved by changing the threshold parameter, the second I do not know.

junghans:
Could we use the same comparison as here:
https://github.com/votca/csg/blob/master/share/scripts/inverse/functions_common.sh#L745
i.e.:
```
if (x==y){
  return equal;
}else{
  if (abs(x-y)<error){
    return equal;
  }else{ 
    return ( abs(x-y)/max(abs(x),abs(y)) < e );
  }
}
```

JensWehner:
my commit an that should be the same, I think

JensWehner:
I check it tomorrow with more care.

junghans:
Ping @JensWehner

JensWehner:
yeah I know. The situation has not changed.

junghans:
@JoshuaSBrown please review.

JensWehner:
@junghans thanks for fixing this finally.

PR #144, opened 2019-05-04
Added tests for unitconverter

JoshuaSBrown:
Class for helping the io readers in csg convert between different units.

### DIFF
--- a/include/votca/tools/unitconverter.h
+++ b/include/votca/tools/unitconverter.h
@@ -94,6 +94,7 @@ class UnitConverter {
       case MassUnit::atomic_mass_units:
         return 1.0;
     }
+    return 0.0;
   }
 
   /// All energies in terms of electron volts
--- a/include/votca/tools/unitconverter.h
+++ b/include/votca/tools/unitconverter.h
@@ -44,50 +44,86 @@ enum EnergyUnit { electron_volts, kilocalories, hartrees, joules };
 class UnitConverter {
  private:
   /// All distances with respect to Ang
-  const std::map<DistanceUnit, double> distance_map_ = {
-      {DistanceUnit::meters, 1E10},
-      {DistanceUnit::centimeters, 1E8},
-      {DistanceUnit::nanometers, 10.0},
-      {DistanceUnit::angstroms, 1.0},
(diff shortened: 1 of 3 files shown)


JensWehner:
why do we make it functions and not constexpr or even templated, I mean the stuff is known at compile time. Otherwise not bad, I thought I remember something similar in tools somewhere

JoshuaSBrown:
I guess I could just return the ratio and not the value into the functions, then I could make it a constexpr. The only thing in tools I see is the constants file, but this is not flexible enough. If I just used the values in constants file it would simply be a hard coded value. If I pass in enums based on the types being used I can at least update the ratio based on the types. E.g. if I'm using your Atom container vs the csg molecule type, if each of the containers has a typedef associated with the units it is using I can simply pass that to the input or output reader during compile time. 

E.g. 

    class AtomContainer {
       typedef tools::DistanceUnit::ang distance_units_;
    };

    class Molecule {
        typedef tools::DistanceUnit::nanometer distance_units_;
    };

    /// Assuming default units are nm in the pdbreader
    PDBReader {
         typedef tools::DistanceUnit::nanometer distance_unit_;
         template<class T> 
         void Read(T & container){
             :
             double x_pos << file_identifier;
             Converter converter;
             double x_pos_correct_units = converter.convert(distance_unit_,T::distance_unit_) * x_pos;
             :
                           
         }
    };

JensWehner:
`double x_pos_correct_units = converter.convert<distance_unit_,T::distance_unit_>()*x

JensWehner:
otherwise it is a great idea, which could solve all unit problems. That is why I am commenting..

JoshuaSBrown:
Hmm, how would you enforce that the compatible enums are passed in. Template specialization? 

E.g. Preventing the following which does not make sense:

    convert<distance_unit_,T::time_unit_>();

JensWehner:
Ohh good point, your way we would get a compile error because no overload an be found, or not?

We do a std::enable_if(std::same(INPUT,length)) or so i am not sure it works fully yet

JoshuaSBrown:
@JensWehner So I made some changes to make it run at compile time, if you don't have any issues, I'll just merge this how it is.

PR #160, opened 2019-09-16
changed threading handling

JensWehner:
Changed the default handling


CMakeModules/FindMKL.cmake
```diff
@@ -278,3 +283,5 @@ if(MKL_FOUND)
         IMPORTED_IMPLIB "${MKL_Core_LINK_LIBRARY}")
   endif()
 endif()
+
+message("-- " ${MKL_THREAD_LAYER} " used for MKL parallelisation")
```
junghans:
```suggestion
message(STATUS "'${MKL_THREAD_LAYER}' used for MKL parallelisation")
```

CMakeModules/FindMKL.cmake
```diff
@@ -62,7 +62,12 @@ include(CheckTypeSize)
 check_type_size("int" INT_SIZE
   BUILTIN_TYPES_ONLY LANGUAGE C)
 
-set(MKL_THREAD_LAYER "Sequential" CACHE STRING "The thread layer to choose for MKL")
+
+if(NOT MKL_THREAD_LAYER)
+    set(MKL_THREAD_LAYER "TBB")
+endif()
+
+set(MKL_THREAD_LAYER ${MKL_THREAD_LAYER} CACHE STRING "The thread layer to choose for MKL")
```
junghans:
```suggestion
set(MKL_THREAD_LAYER "TBB" CACHE STRING "The thread layer to choose for MKL")
```
User's choice will always overwrite.

CMakeModules/FindMKL.cmake
```diff
@@ -62,7 +62,12 @@ include(CheckTypeSize)
 check_type_size("int" INT_SIZE
   BUILTIN_TYPES_ONLY LANGUAGE C)
 
-set(MKL_THREAD_LAYER "Sequential" CACHE STRING "The thread layer to choose for MKL")
+
+if(NOT MKL_THREAD_LAYER)
+    set(MKL_THREAD_LAYER "TBB")
+endif()
```
junghans: Drop these 3 lines.

CMakeModules/FindMKL.cmake
JensWehner: Why did it not work before then?

JensWehner: That line was exactly my problem.

junghans: Before the default was `Sequential`, right?

JensWehner: Yes but in tools CMakeList the parameter was explicitly set

JensWehner: But FindMKL did not use it

junghans: I see, `set(MKL_THREAD_LAYER "Intel OpenMP")` isn't enough, you need to do something like `set(MKL_THREAD_LAYER "Intel OpenMP" CACHE STRING ...)` there.

PR #163, opened 2019-09-18
fix includes

JensWehner:
removes all using namespace  from include headers, merge after https://github.com/votca/csg/pull/409

### DIFF
--- a/src/libtools/histogram.cc
+++ b/src/libtools/histogram.cc
@@ -38,8 +38,8 @@ void Histogram::ProcessData(DataCollection<double>::selection *data) {
   _pdf.assign(_options._n, 0);
 
   if (_options._auto_interval) {
-    _min = numeric_limits<double>::max();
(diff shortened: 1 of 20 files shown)


src/libtools/histogram.cc
```diff
@@ -38,8 +38,8 @@ void Histogram::ProcessData(DataCollection<double>::selection *data) {
   _pdf.assign(_options._n, 0);
 
   if (_options._auto_interval) {
-    _min = numeric_limits<double>::max();
-    _max = numeric_limits<double>::min();
+    _min = std::numeric_limits<double>::max();
```
JoshuaSBrown: Any particular reason why you are using std in the source files, is it not safe to simply do a using namespace std?

JensWehner: in general I would not use using namespace anywhere. So when it is just a couple std:: I just add them

PR #169, opened 2019-10-03
enable eigen tensor module

JensWehner:
This serves to replace boost multiarray. Eigen::Tensor is now used by Tensorflow, i.e. supported by Google. I think the support is good enough now to warrant deprecating boost::multiarray.


include/votca/tools/eigen.h
```diff
@@ -44,6 +44,7 @@
 #pragma GCC diagnostic ignored "-Wint-in-bool-context"
 #endif
 #include <Eigen/Eigen>
+#include <unsupported/Eigen/CXX11/Tensor>
```
junghans: Is it really in a folder `unsupported/` ?

JensWehner: yeah. It is an unofficial module of eigen but since google uses it in tensorflow. I think it is good enough

junghans: ok

JensWehner: https://stackoverflow.com/questions/41518379/why-was-eigen-chosen-for-tensorflow

PR #170, opened 2019-10-05
replaced random2 internals with std::random

### LINKED ISSUE #118: Replace random2 with std::random implementation

JensWehner:
https://github.com/votca/tools/issues/118  and https://github.com/votca/xtp/pull/293

Fix #118


src/tests/test_random2.cc
```diff
@@ -0,0 +1,69 @@
...
+  int seed = 1;
+  random.init(seed);
+  std::vector<double> results;
+  int number = 1e5;
+  results.reserve(number);
+  for (int i = 0; i < number; i++) {
+    results.push_back(random.rand_uniform());
+  }
+
+  // average should be close to 25
```
junghans: 0.5 I guess?

JensWehner: yep

junghans:
2 questions: why aren’t we naming it “Random” without the “2”.  And why do we need this class at all?

JensWehner:
Good questions. a) history, we can rename it, but I was to lazy. We should probably rename it. b) We could also not use it, I guess this gives it an even easier interface.

junghans:
Then, I would say let's switch xtp to `std::random` and kill this class with fire.

JensWehner:
If we delete the class here I would just move it to xtp, because I need exactly this interface. I do not care it is up to you.

junghans:
Then, let's leave it here, but rename is to `Random`.

JensWehner:
okay. I ll do that

PR #172, opened 2019-10-12
[WIP] Blas integration

junghans:


CMakeModules/FindMKL.cmake
```diff
@@ -84,10 +84,10 @@ find_path(MKL_INCLUDE_DIR
     mkl_blas.h
     mkl_cblas.h
   PATHS
-    /opt/intel
-    /opt/intel/mkl
+    #/opt/intel
+    #/opt/intel/mkl
```
junghans: Why?

junghans:
Maybe it would preferable to have a Eigen backend, e.g. `EIGEN_BACKEND`, as mkl and blas seem to be mutual exclusive.

junghans:
@JensWehner do we use anything beside blas routines out of mkl? If not we can simply detect mkl as a blas provider and do not special handling of it at all.

junghans:
With `-DBLA_VENDOR=Intel`

JensWehner:
Yes, the whole LAPACK and the MKL math library and some sparse stuff and maybe at some point the fft. So even if we add LAPACK, MKL will always be a superset.

JensWehner:
We also have to find out, which vendors we provided, because openmp does not interact with all of them the same way.

junghans:
There might be a CMake variable which tells you the vendor.

junghans:
Looks ready to me, @JensWehner your call!

junghans:
> a) we have to parse the vendors through

One of my systems has netlib blas and no vendor string in the library, so that is no going to be easy.

JensWehner:
We need at least for the libraries which have their own OPENMP thread directives, because otherwise there will be threads all over the place.

JensWehner:
and at the moment unless you have a really optimized library it will either be slower or just about as good as pure eigen. So at the moment I cannot see the need.

baumeier:
Eigen vs. openblas vs. cuda:
https://medium.com/datathings/benchmarking-blas-libraries-b57fb1c6dc7

Seems to indicate that for most of what we do now, the Eigen internal calculations are surprisingly good. Only very few cases in which openblas is really better, e.g. matrix multiplication with matrices bigger than 900. It definitely requires more testing to see if it is worth it, but again that could be specific to a system.

Maybe we should keep this as an option that is just activated with a Cmake flag -DWITH_EXTERNALBLAS=ON or something?

JensWehner:
I would only merge it if the openmp stuff is sorted out

junghans:
I made this off by default, so we we could merge it now.

One question, what does Eigen do if `EIGEN_USE_MKL_ALL` and `EIGEN_USE_BLAS` are used?

JensWehner:
@junghans I have no idea what happens if both are on. I am not a friend of merging it now, as we have no experience with it nor a testcase, nor is it used anywhere. It also seems only MKL atm justifies the extra complications, because with other blas libraries the performance improvement was minimal at best.

PR #182, opened 2019-11-10
Fft rewrite

### LINKED ISSUE #181: Replace fftw interface with Eigen fftw
would allow us to use eigen internal implementation without fftw library 
and otherwise we could use fftw or MKL as a backend. 

https://eigen.tuxfamily.org/dox/unsupported/group__FFT__Module.html

https://gitlab.math.ethz.ch/NumCSE/NumCSE/blob/43c3152296a97f4a927cd82fe3c19ac0c64537e4/CMake/Macros/set_eigen_fft_backend.cmake

JensWehner:
implements https://github.com/votca/tools/issues/181

Part of https://github.com/votca/votca/pull/219

Fix #181 
Related to https://github.com/votca/csg/issues/104

### DIFF
--- a/include/votca/tools/eigen.h
+++ b/include/votca/tools/eigen.h
@@ -28,7 +28,6 @@
 #endif
 
 #ifndef NOFFTW
-#include <fftw3.h>
 #define EIGEN_FFTW_DEFAULT
 #endif
 
--- a/include/votca/tools/eigen.h
+++ b/include/votca/tools/eigen.h
@@ -27,6 +27,11 @@
 #define EIGEN_USE_MKL_ALL
 #endif
 
+#ifndef NOFFTW
+#include <fftw3.h>
+#define EIGEN_FFTW_DEFAULT
+#endif
+
 //! Macro to detect strictly gcc.
 //! \details __GNUC__ and __GNUG__ were intended to indicate the GNU compilers.
 //! However, they're also defined by Clang/LLVM and Intel compilers to indicate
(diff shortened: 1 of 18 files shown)


include/votca/tools/eigen.h
```diff
@@ -27,6 +27,11 @@
 #define EIGEN_USE_MKL_ALL
 #endif
 
+#ifndef NOFFTW
+#include <fftw3.h>
```
junghans: I don't think this line is needed as `/usr/include/eigen3/unsupported/Eigen/FFT` includes `fftw3.h` already.

JensWehner: ahh good point.

include/votca/tools/eigen.h
```diff
@@ -45,6 +49,7 @@
 #endif
 #include <Eigen/Eigen>
 #include <unsupported/Eigen/CXX11/Tensor>
+#include <unsupported/Eigen/FFT>
```
junghans: Is there a reason to always include this? Why not just the correlate class?

JensWehner: I thought it is maybe a good idea to make it available to all of votca, as people might want to use this functionality.

JensWehner: because FFT could come up quite quickly in XTP as well

JensWehner: but putting it into correlate would probably be enough

JensWehner: I think FFT could be useful for further development and it is pretty well namespaced so I actually would make it available in all of votca

JensWehner: what do you think?

junghans: No preference just wondering.

PR #211, opened 2020-01-13
cmake: remove old html target, now done in votca/votca

junghans:


junghans:
> is there a PR in votca/votca? Which lines correspond to it?

It is the `doxygen` Target in votca/votca.

PR #225, opened 2020-04-15
Stable fixes

junghans:


JensWehner:
Can you briefly explain what the problem was?

junghans:
> Can you briefly explain what the problem was?

Installing our `FindMKL.cmake` and using it, is just the cleaner solution than recreating the imported target in the exported target.

PR #227, opened 2020-04-17
fix merging calculator defaults

### LINKED ISSUE #226: Error when merging calculator defaults with User input
The [Calculator's method UpdateWithUserOption](https://github.com/votca/tools/blob/master/include/votca/tools/calculator.h#L93), when merging the `user_options` with the `defaults` for a given calculator, assumes that the `defaults` options always contains the same or more information than the `user_options`, but there are *options* without a reasonable default that must be passed by the user.
This case has not been taken into account when merging.

felipeZ:
fixed #226


felipeZ:
Converting to Draft, because there is still another bug

felipeZ:
@JensWehner @baumeier  This PR is ready for reviewing

PR #229, opened 2020-04-20
cmake: fix build with mkl

junghans:
Avoid an error when calling find_package(MKL) twice in xtp during
detecting tools and csg. Also remember mkl thread settings in
exported target and do not pollute CMAKE_MODULE_PATH.


junghans:
> I do not fully understand it.

The problem is xtp looks for csg and tools and csg looks for tools again hence `find_package(MKL)` gets called twice and CMake fails as the `MKL::MKL` target already exists.

PR #230, opened 2020-04-20
cmake: fix usage of CMAKE_MODULE_PATH (could be a list)

junghans:


junghans:
@JensWehner please review again!

PR #233, opened 2020-04-24
check calculator choices

### LINKED ISSUE #232: Check that the calculator input option has a valid input
## Motivation
* Unify all the defaults inside the `xml` files.
* Reduce the amount of check performs by each `Calculator.Initialize` method.

## Proposed changes
* Read a new `choices` attribute for each option in the  XML `Calculators` file.
* Check that the value of each option after merging with the *user input* has valid input.
* Introduced 5 new choices for continuous values: `[bool,float, float+, int, int+]` to represent reals, positive reals, integers, and positive integers, respectively.
 
## Example
suppose there is a calculator with the following defaults:
```xml
<options>
  <calculator>
       <opt1 help="first option" default="foo" choices="foo,bar">foo</opt1>
       <opt2 help="second option" default="3.1415" choices="float+">3.1415</opt2> 
 </calculator>
</options>
```
The user can change both `opt1` and `opt2` but the changes must be valid values as declared by `choices`

### Note
The options without a `choices` attribute is skip.

felipeZ:
fix #232


include/votca/tools/calculator.h
```diff
@@ -23,6 +23,7 @@
 #include "globals.h"
 #include "property.h"
 #include "propertyiomanipulator.h"
+#include "tokenizer.h"
```
JensWehner, line 26: you only need this in the .cc file?

src/libtools/calculator.cc
```diff
@@ -75,5 +75,66 @@ void Calculator::OverwriteDefaultsWithUserInput(const Property &p,
...
+        }
+      }
+    }
+  }
+}
+
+void Calculator::CheckOption(const Property &prop,
+                             const std::vector<std::string> &choices) {
+  const std::string &head = choices.front();
+  if (head == "bool") {
```
JensWehner, line 116: what happens if the input cannot be converted to bool?

src/tests/test_calculator.cc
```diff
@@ -46,8 +46,9 @@ BOOST_AUTO_TEST_CASE(load_defaults_test) {
       std::ofstream defaults("calculators/xml/testcalc.xml");
       defaults << "<options>\n"
                << "<testcalc>\n"
-               << "<option1>0</option1>\n"
-               << "<option2>3.141592</option2>\n"
+               << "<option0 choices=\"foo,bar\">foo</option0>\n"
```
JensWehner, line 49: we need also test cases for the error cases.

src/libtools/calculator.cc
felipeZ: If it cannot convert the input to `bool`, `float` and `int` it raises an error. I am going to add a test for it :)

include/votca/tools/calculator.h
felipeZ: I am moving it to the `.cc` thanks!

src/tests/test_calculator.cc
felipeZ: See next commit

src/libtools/calculator.cc
JensWehner: yeah but the error is not very descriptive is it? Btw. it is weekend. :)

src/libtools/calculator.cc
```diff
@@ -75,5 +76,75 @@ void Calculator::OverwriteDefaultsWithUserInput(const Property &p,
...
+  const std::string &head = choices.front();
+  std::ostringstream oss;
+
+  if (head == "bool") {
+    prop.as<bool>();
+  } else if (head == "float") {
+    prop.as<double>();
+  } else if (head == "float+") {
+    if (prop.as<double>() < 0.0) {
+      throw std::runtime_error("");
```
JensWehner, line 132: should these errors not be more verbose? i.e. `option "a" needs a float, your argument "b" cannot be converted to it`

felipeZ: That is done above in line 98: https://github.com/votca/tools/blob/choices/src/libtools/calculator.cc#L98
That runtime error is caught there and verbose error is printed. For example, for the [test](https://github.com/votca/tools/blob/choices/src/tests/test_calculator.cc#L95) something like this is printed:
```batch
2: The input value for "option1" should be one of the following values: "foo" "bar" "baz" "qux"  But "boom" cannot be converted into one.
2: The input value for "option2" should be a "float" But "some" cannot be converted into one.
2: The input value for "option3" should be a "int" But "3.14" cannot be converted into one.
2: The input value for "option4" should be a "int+" But "-2" cannot be converted into one.
2: The input value for "option5" should be a "float+" But "-3.14" cannot be converted into one.
```

PR #234, opened 2020-04-25
fix build with non-system fftw

junghans:
Non-system fftw will lead to a missing 'fftw.h' when building
excutables as fftw is only a private dependcies and hence the
fftw include path is missing. Don't include fft header in
eigen.h, so it won't get pull in for every executable.

Reported here:
https://groups.google.com/d/msg/votca/wFNG1mmKwSU/hOS1zplRDQAJ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ For more detailed information about the changes see the history of the [reposito
 
 ## Version 1.6.1 (released XX.04.20)
 * fix build with mkl (#229)
+* fix build with non-system libfftw (#234)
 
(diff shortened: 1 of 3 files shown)


junghans:
@JensWehner we had this discussion before https://github.com/votca/tools/pull/182#discussion_r344520822

JensWehner:
Yeah I still think fft could be useful for the rest but fixing the bug is more important and we can still change it when it comes up

junghans:
> Yeah I still think fft could be useful for the rest but fixing the bug is more important and we can still change it when it comes up

Yeah, let’s change it, when we use it somewhere outside of the correlate class.

PR #235, opened 2020-04-26
cmake: make fftw3 public

junghans:
@jens this is the "real" fix for #234


junghans:
@jens, do you like this solution better?

PR #236, opened 2020-04-26
Cmake clean up

junghans:


junghans:
Need to merge https://github.com/votca/votca/pull/280 first.

PR #240, opened 2020-05-06
CI: sync changes from votca/votca

junghans:


junghans:
I just want to test both.

On a single full compile ninja isn't much faster, but it is usually faster for incremental builds.

PR #242, opened 2020-05-08
update option only if new value is not empty

felipeZ:
## Fixed bug 
Following the new standard proposed in #241, only the `default` attribute contains information and the values are empty for the default `XML` files. When merging the `default` options with the `user` options, some of the user's values may be empty, resulting in a `Property` object that contains some empty values even though they have a well-defined default.

### Proposed changes
Only overwrite the default values if the `user input` is non-empty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ For more detailed information about the changes see the history of the [reposito
 * check calculator input options (#232, #233)
 * allow calculator choices to be a list (#239)
 * inject defaults into calculator values (#241)
+* fixed bug introduced by defaults injection (#241)
 
 ## Version 1.6.1 (released XX.04.20)
 * fix build with mkl (#229)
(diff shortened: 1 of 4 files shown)


felipeZ:
@JensWehner I have added these [2 lines of code](https://github.com/votca/tools/blob/19c4e389dee8d8cc917e90e1149d45c7d9731e19/include/votca/tools/calculator.h#L105) that looks useless, but they cover the corner case the following corner case:
We are merging the defaults with the user input, for a given `option` there is no default property but there is a `user option` for which the value is empty, but the `user option` itself has a default.

Yes, you are going to tell me *why the user doesn't fill in the values?*.  The answer is that sometimes the user just grabs those values from another XML, that can happen for instance in the **qmmm** calculations where the user *copy/paste* the dft, gwbse, etc. options from the XML files.

PR #245, opened 2020-05-21
votca_compare: add +/-'nan' comparision

junghans:


junghans:
> I still do not get why we have nan values in the tests. Otherwise it looks fine

It is in third (error) column and for very short runs that we are doing during the regression tests, the error is sometimes nan as the error is `sqrt((sum2-sum*sum/c)/(c*(c-1)))` and for `c=1`, you get `0/0` and hence `nan`.

junghans:
I have another idea how to get rid of the `nan` so let's hold off on this for a bit.

junghans:
Not needed anymore.

PR #248, opened 2020-05-30
For/stable/joshua s brown fix abs

JoshuaSBrown:
The wrong abs function was being used. Somewhere a c header was being included that was converting a double to int when calling abs. This was mitigated by using fabs.

The bug was discovered as a result in the lammps data reader in csg: https://github.com/votca/csg/issues/537

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ For more detailed information about the changes see the history of the [reposito
 * fix build with mkl (#229)
 * fix build with non-system libfftw (#234)
 * fix CI on Ubuntu-20.04 (#237)
+* fix bug related to calling c abs instead of c++ fabs (#248)
 
 ## Version 1.6 _SuperPelagia_ (released 17.04.20)
 * fix clang-10 warnings (#217)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ For more detailed information about the changes see the history of the [reposito
 * fix build with non-system libfftw (#234)
 * fix CI on Ubuntu-20.04 (#237)
 * fix bug related to calling c abs instead of c++ fabs (#248)
+* updated floating point comparison to use boost epsilon error (#248)
+* promoted boost version from 1.53 to 1.63 (#248)
 
 ## Version 1.6 _SuperPelagia_ (released 17.04.20)
 * fix clang-10 warnings (#217)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ For more detailed information about the changes see the history of the [reposito
 * fix bug related to calling c abs instead of c++ fabs (#248)
 * updated floating point comparison to use boost epsilon error (#248)
 * promoted boost version from 1.53 to 1.63 (#248)
+* fix compile error in structure parameters by adding hash function (#248)
 
 ## Version 1.6 _SuperPelagia_ (released 17.04.20)
 * fix clang-10 warnings (#217)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,7 @@ For more detailed information about the changes see the history of the [reposito
 * fix build with non-system libfftw (#234)
 * fix CI on Ubuntu-20.04 (#237)
 * fix bug related to calling c abs instead of c++ fabs (#248)
-* updated floating point comparison to use boost epsilon error (#248)
-* promoted boost version from 1.53 to 1.63 (#248)
+* updated floating point comparison in akimaspline.h (#248)
 * fix compile error in structure parameters by adding hash function (#248)
 
 ## Version 1.6 _SuperPelagia_ (released 17.04.20)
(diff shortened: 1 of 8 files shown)


include/votca/tools/akimaspline.h
```diff
@@ -98,8 +98,8 @@ inline double AkimaSpline::getSlope(double m1, double m2, double m3,
   if ((m1 == m2) && (m3 == m4)) {
```
junghans, line 98: We should fix this comparison, too.

JoshuaSBrown: Wow you are fast. I just created this.

JoshuaSBrown: And what would be a suitable tolerance? I have no idea what sizes these numbers typically have.

JensWehner:
Heyho I am not convinced. I do not see how that should happen, std::abs is templated.

junghans:
I think replacing `abs()` by `std::abs()` should be enough.

JoshuaSBrown:
Ok, std::abs then, we should just be careful when using abs though. @JensWehner which part are you not convinced about?

junghans:
And can we add a test to csg, comparing the output of csg_dump.

JensWehner:
works for me

JensWehner:
we should try to remove all using namespace std;

JoshuaSBrown:
> And can we add a test to csg, comparing the output of csg_dump.

I don't think this is necessary, this errors is actually caught in the tools unit element test as it is.

CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ endif(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/.git)
 
 find_package(Threads REQUIRED)
 
-find_package(Boost 1.53.0 REQUIRED COMPONENTS program_options filesystem system )
+find_package(Boost 1.63.0 REQUIRED COMPONENTS program_options filesystem system )
```
junghans, line 57: I think that might be a bad idea, we still need to support bionic for @baumeier, which only has boost-1.65: https://packages.ubuntu.com/bionic/libboost-all-dev

JoshuaSBrown: Ok

src/libtools/histogram.cc
```diff
@@ -76,7 +76,7 @@ void Histogram::ProcessData(DataCollection<double>::selection* data) {
   if (_options._scale == "bond") {
     for (size_t i = 0; i < _pdf.size(); ++i) {
       double r = _min + _interval * (double)i;
-      if (abs(r) < 1e-10) {
+      if (fabs(r) < 1e-10) {
```
junghans, line 79: `std::` here?

src/libtools/histogram.cc
```diff
@@ -86,7 +86,7 @@ void Histogram::ProcessData(DataCollection<double>::selection* data) {
     for (size_t i = 0; i < _pdf.size(); ++i) {
       double alpha = _min + _interval * (double)i;
       double sa = sin(alpha);
-      if (abs(sa) < 1e-5) {
+      if (fabs(sa) < 1e-5) {
```
junghans, line 89: `std::` here?

JensWehner:
I think we should not allows using namspace std;. CLang-tidy flags it as well.

junghans:
> I think we should not allows using namspace std;. CLang-tidy flags it as well.

This is for stable, let's fix it in master.

src/libtools/histogram.cc
JoshuaSBrown: done

src/libtools/histogram.cc
JoshuaSBrown: done

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ For more detailed information about the changes see the history of the [reposito
 * fix build with mkl (#229)
 * fix build with non-system libfftw (#234)
 * fix CI on Ubuntu-20.04 (#237)
+* fix bug related to calling c abs instead of c++ fabs (#248)
+* updated floating point comparison to use boost epsilon error (#248)
+* promoted boost version from 1.53 to 1.63 (#248)
```
junghans, line 9: This can go!

include/votca/tools/structureparameters.h
```diff
@@ -105,7 +105,7 @@ class StructureParameters {
   T get(const StructureParameter parameter) const;
 
  private:
-  std::unordered_map<StructureParameter, boost::any> parameters;
+  std::unordered_map<StructureParameter, boost::any, std::hash<int>> parameters;
```
junghans, line 108: Why is that needed?

CHANGELOG.md
JoshuaSBrown: Shoot.

include/votca/tools/structureparameters.h
JoshuaSBrown: It turns out tools was also broken on ubuntu 16 in an unrelated issue. This fixes that.

junghans: Ok, Ubuntu 16 is pretty old, correct? Was the boost bump related?

JoshuaSBrown: Um no, that was for the floating point comparison issue in akimaspline. I wanted to use the boost epsilon which is only available starting in boost 1.63

JoshuaSBrown: Yes, 16 has been out for a while.

PR #254, opened 2020-06-18
GH Actions: add format action

junghans:
This won't work before actions got added to master.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -7,7 +7,7 @@ jobs:
     name: clang-format
     if: ${{ github.event.comment.body == '@votca-bot format' }}
     runs-on: ubuntu-latest
-    container: votca/buildenv:latest
+    container: votca/buildenv:format
     steps:
       - name: Check if message comes from PR
         uses: octokit/request-action@v2.x
@@ -34,12 +34,12 @@ jobs:
           ref: ${{ fromJSON(steps.pr_info.outputs.data).head.ref }}
       - name: Run clang-format
         if: ${{ fromJSON(steps.issue_info.outputs.data).pull_request }}
-        run: find . -type f -regex ".*\.\(cc\|h\)" -print0 | xargs -0 clang-format -i -style=file
+        run: find . -type f -regex ".*\.\(c\|h\|cc\|cpp\|hpp)" -print0 | xargs -0 clang-format -i -style=file
       - name: Commit and push
         if: ${{ fromJSON(steps.issue_info.outputs.data).pull_request }}
         run: |
           git config --global user.name "Votca Bot"
           git config --global user.email "github@votca.org"
           git add -u
-          git commit -m "Format code using clang-format" || true
+          git commit -m "Format code using $(clang-format --version)" || true
           git push
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,45 @@
+name: Format
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  edit:
+    name: clang-format
+    if: ${{ github.event.comment.body == '@votca-bot format' }}
+    runs-on: ubuntu-latest
+    container: votca/buildenv:latest
+    steps:
+      - name: Check if message comes from PR
+        uses: octokit/request-action@v2.x
+        id: issue_info
+        with:
+          route: GET /repos/:repository/issues/:pull_number
+          repository: ${{ github.repository }}
+          pull_number: ${{ github.event.issue.number }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Get PR info
+        if: ${{ fromJSON(steps.issue_info.outputs.data).pull_request }}
+        uses: octokit/request-action@v2.x
+        id: pr_info
+        with:
+          route: GET /repos/:repository/pulls/:pull_number
+          repository: ${{ github.repository }}
+          pull_number: ${{ github.event.issue.number }}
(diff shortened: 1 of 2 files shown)


.github/workflows/format.yml
```diff
@@ -0,0 +1,45 @@
+name: Format
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  edit:
+    name: clang-format
+    if: ${{ github.event.comment.body == '@votca-bot format' }}
```
junghans, line 8: Do we want a different magic phrase? Or should be `startWith()` here instead of exact match?

.github/workflows/format.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    container: votca/buildenv:latest
```
junghans, line 10: I don't like the fact that pulling the container takes 60sec out of the 62sec runtime. 
Maybe we should make a small container, which has nothing but clang-format in there.

.github/workflows/format.yml
```diff
@@ -0,0 +1,45 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Checkout head branch from PR
+        if: ${{ fromJSON(steps.issue_info.outputs.data).pull_request }}
+        uses: actions/checkout@master
+        with:
+          repository: ${{ fromJSON(steps.pr_info.outputs.data).head.repo.full_name }}
+          ref: ${{ fromJSON(steps.pr_info.outputs.data).head.ref }}
+      - name: Run clang-format
+        if: ${{ fromJSON(steps.issue_info.outputs.data).pull_request }}
+        run: find . -type f -regex ".*\.\(cc\|h\)" -print0 | xargs -0 clang-format -i -style=file
```
junghans, line 37: I decided against running `make format` here, so that we don't have to call `votca/action/setup` and `cmake` first.
However, if we ever change file extensions this will break.

.github/workflows/format.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Commit and push
+        if: ${{ fromJSON(steps.issue_info.outputs.data).pull_request }}
+        run: |
+          git config --global user.name "Votca Bot"
+          git config --global user.email "github@votca.org"
+          git add -u
+          git commit -m "Format code using clang-format" || true
+          git push
```
junghans, line 45: We could refactor these 5 lines in a custom action like `votca/actions/setup`, but that seemed a bit over the top.

.github/workflows/format.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Run clang-format
+        if: ${{ fromJSON(steps.issue_info.outputs.data).pull_request }}
+        run: find . -type f -regex ".*\.\(cc\|h\)" -print0 | xargs -0 clang-format -i -style=file
+      - name: Commit and push
+        if: ${{ fromJSON(steps.issue_info.outputs.data).pull_request }}
+        run: |
+          git config --global user.name "Votca Bot"
+          git config --global user.email "github@votca.org"
+          git add -u
+          git commit -m "Format code using clang-format" || true
```
junghans, line 44: Commit message ok?

junghans:
@JensWehner @JoshuaSBrown please have a look at my comment above and discuss.

.github/workflows/format.yml
JoshuaSBrown: I think this is fine. It is simple and clear.

.github/workflows/format.yml
JoshuaSBrown: If it's not a hassle I'm in agreement.

.github/workflows/format.yml
JoshuaSBrown: Why not just go ahead and add the file extensions that we might possible switch too. (.hpp/.cpp)

.github/workflows/format.yml
JoshuaSBrown: Can we do "votca-bot: Format code using clang-format"

.github/workflows/format.yml
JoshuaSBrown: I think it is fine.

.github/workflows/format.yml
junghans: Good point, will do!

.github/workflows/format.yml
junghans: @votca-bot is already the author of the commit.  Including the version seems reasonable.

.github/workflows/format.yml
junghans: It is easy.

.github/workflows/format.yml
JensWehner: uih nice

.github/workflows/format.yml
JensWehner: I agree

.github/workflows/format.yml
JensWehner: good points.

JensWehner:
I really like this :)

.github/workflows/format.yml
junghans: https://github.com/votca/buildenv/pull/86, please review

junghans:
@jens I made the action not VOTCA-specific so you can reuse it for other projects.

.github/workflows/format.yml
junghans: It will say "Format code using clang-format version 10.0.0 (Fedora 10.0.0-2.fc32)" now.

PR #266, opened 2020-06-29
Fix FindMKL.cmake

junghans:


junghans:
Hmm, this was supposed to be a reproducer for votca/votca#350, but it doesn't fail here either, not even in `votca/buildenv:fedora_gmx2020_d`

junghans:
See votca/xtp#493

PR #273, opened 2020-07-04
cmake: enable CXX only

junghans:


JoshuaSBrown:
Why are the checks no longer needed? Or does specifying the language inherently check that the correct files are available?

junghans:
> Why are the checks no longer needed? Or does specifying the language inherently check that the correct files are available?

- AIX is not very common anymore, we haven't build on it for years.
- we don't use `assert.h` anymore, but `cassert` instead
- we are now using `std::sqrt()` instead of `sqrt()`, so the check is overly strict.

PR #278, opened 2020-07-12
format code with clang-10

junghans:
Need this change as the base container is now F32.


junghans:
@votca-bot format

JensWehner:
do you want clang-format 10 on all repositories? In my experience now, it is much better to lock the clang-format level to one specific version, because they cannot keep it backwards compatible.

junghans:
Fedora only has one clang-format Version installed, plus this is the only change needed in all repos.

PR #280, opened 2020-07-14
Move to GitHub Actions

junghans:


junghans:
@JensWehner do you think we need coverage tracking for the `stable` branch?

I also want to create an action that creates a PR from `stable` to `master` automatically.

PR #309, opened 2020-11-16
strip windows line endings in readers

junghans:


JensWehner (approved):
Thanks, we should do the same for xtp as well. I can do that later unless you have your sed script ready already.

junghans:
@JensWehner see https://github.com/votca/xtp/pull/596

PR #313, opened 2020-12-15
add option to print man page in rst format

### LINKED ISSUE #312: generate rst page from help
The `--man` option outputs a manpage, could we add something similar to print out an rst page to include in the website?

felipeZ:
## Change
* Output man pages in `rst` format
* Add option to write man pages to file

fix #312

### DIFF
--- a/src/libtools/application.cc
+++ b/src/libtools/application.cc
@@ -62,16 +62,13 @@ void Application::ShowManPage(std::ostream &out) {
   out << FormatManPage<globals::man>();
 }
 
-void Application::WriteManFile() {
-  ofstream man_file;
-  man_file.open(ProgramName() + "_man.rst");
-  man_file << FormatManPage<globals::manrst>();
-  man_file.close();
+void Application::ShowRstPage(std::ostream &out) {
+  out << FormatManPage<globals::manrst>();
 }
 
 template <typename Format>
 std::string Application::FormatManPage() {
-  std::string version = (VersionString() != "") ? VersionString() : "1.0";
+  std::string version = ToolsVersionStr();
   std::stringstream stream;
   stream << boost::format(Format::header) % ProgramName() % version;
   stream << boost::format(Format::name) % ProgramName() % globals::url;
@@ -121,8 +118,7 @@ int Application::Exec(int argc, char **argv) {
     AddProgramOptions()("verbose1", "  be very loud and noisy");
     AddProgramOptions()("verbose2,v", "  be extremly loud and noisy");
     AddProgramOptions("Hidden")("man", "  output man-formatted manual pages");
-    AddProgramOptions("Hidden")(
-        "manrst", " write the manual pages to a rst formatted file ");
+    AddProgramOptions("Hidden")("rst", "  output rst-formatted manual pages ");
     AddProgramOptions("Hidden")("tex", "  output tex-formatted manual pages");
 
     Initialize();  // initialize program-specific parameters
@@ -147,8 +143,8 @@ int Application::Exec(int argc, char **argv) {
       return 0;
     }
 
-    if (_op_vm.count("manrst")) {
-      WriteManFile();
+    if (_op_vm.count("rst")) {
+      ShowRstPage(cout);
       return 0;
     }
 
--- a/src/libtools/application.cc
+++ b/src/libtools/application.cc
@@ -84,7 +84,8 @@ std::string Application::FormatManPage() {
   for (const auto &option : _op_desc.options()) {
     string format_name = option->format_name() + option->format_parameter();
     boost::replace_all(format_name, "-", "\\-");
-    stream << boost::format(Format::option) % format_name % option->description();
+    stream << boost::format(Format::option) % format_name %
+                  option->description();
   }
 
   stream << boost::format(Format::authors) % globals::email;
--- a/src/libtools/application.cc
+++ b/src/libtools/application.cc
@@ -58,26 +58,33 @@ void Application::ShowHelpText(std::ostream &out) {
   out << "\n\n" << VisibleOptions() << endl;
 }
 
-void Application::ShowManPage(std::ostream &out) {
-
-  out << boost::format(globals::man::header) % ProgramName() % VersionString();
-  out << boost::format(globals::man::name) % ProgramName() % globals::url;
-  out << boost::format(globals::man::synopsis) % ProgramName();
+std::string Application::ShowManPage() {
+  std::string version = (VersionString() != "") ? VersionString() : "1.0";
+  std::stringstream stream;
+  stream << boost::format(globals::man::header) % ProgramName() % version;
+  stream << boost::format(globals::man::name) % ProgramName() % globals::url;
+  stream << boost::format(globals::man::synopsis) % ProgramName();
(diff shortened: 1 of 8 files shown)


JensWehner (commented):
Thanks for doing this @felipeZ . Really quick work.

Does this retain the old functionality completely?

We should also add a test I think.

Is `manfile` a good name, I thought we just call it `rst`?

felipeZ:
> Thanks for doing this @felipeZ . Really quick work.
> 
> Does this retain the old functionality completely?
> 
> We should also add a test I think.
> 
> Is `manfile` a good name, I thought we just call it `rst`?

@JensWehner I have replaced the on `groff` format by `rst`. @junghans  It that is not the intended behaviour I can create a separated function to write down `rst` and another one to print `groff`

JensWehner:
I am not sure, but I think the man pages are supposed to remain @junghans

junghans:
Yes I would to keep both!

I want to keep the main page, but also publish the content on the website too.

felipeZ:
@JensWehner  what do you think? :)

src/libtools/application.cc
```diff
@@ -107,6 +120,8 @@ int Application::Exec(int argc, char **argv) {
     AddProgramOptions()("verbose1", "  be very loud and noisy");
     AddProgramOptions()("verbose2,v", "  be extremly loud and noisy");
     AddProgramOptions("Hidden")("man", "  output man-formatted manual pages");
+    AddProgramOptions("Hidden")(
+        "manrst", " write to a rst formatted file the manual pages");
```
JensWehner, line 124:
```suggestion
        "manrst", " write the manual pages to a rst formatted file ");
```

src/tests/test_application.cc
```diff
@@ -0,0 +1,57 @@
...
+    void HelpText(std::ostream& out) override { out << "Do something awesome"; }
+    void Run() override { ; }
+    void Initialize() override { ; }
+    bool EvaluateOptions() override {
+      WriteManFile();
+      return true;
+    }
+  };
+
+  App app;
```
JensWehner, line 48: aehm why not use std::string?

src/libtools/application.cc
```diff
@@ -59,25 +59,38 @@ void Application::ShowHelpText(std::ostream &out) {
...
+  ofstream man_file;
+  man_file.open(ProgramName() + "_man.rst");
+  man_file << FormatManPage<globals::manrst>();
+  man_file.close();
+}
 
-  out << boost::format(globals::man::header) % ProgramName() % VersionString();
-  out << boost::format(globals::man::name) % ProgramName() % globals::url;
-  out << boost::format(globals::man::synopsis) % ProgramName();
+template <typename T>
```
JensWehner, line 72: What is the idea behind the template parameter?

src/tests/test_application.cc
felipeZ: Because the compiler refuse to convert from `std::string` to `char *`

src/libtools/application.cc
felipeZ: thanks!

src/libtools/application.cc
felipeZ: Printing the manual in `groff`  or writting in `rst` format uses the same functionality but different "data structures". There are now two structures with the data for the manual pages: one in `groff` format and the other in `rst`. They are called `globals::man` and `globals::manrst`, respectively.

src/tests/test_application.cc
JensWehner: ahh does a `.c_str()` work?

src/libtools/application.cc
JensWehner: okay, that makes sense.

JensWehner: but normally `T` is used for generic template parameters, I think in this case we can give a more explicit name to it.

JensWehner: but I am not bent on it. So only do it if you agree

src/tests/test_application.cc
felipeZ: I tried that one too and the compiler doesn't like it either

src/libtools/application.cc
felipeZ: I agree that we can give a more generic type name because is either `man` or `manrst` and both are documentation format. So, what about:
```cc
template <typename Format>
```

JensWehner: nice

JensWehner:
@votca-bot format

src/libtools/application.cc
```diff
@@ -59,25 +59,39 @@ void Application::ShowHelpText(std::ostream &out) {
 }
 
 void Application::ShowManPage(std::ostream &out) {
+  out << FormatManPage<globals::man>();
+}
+
+void Application::WriteManFile() {
+  ofstream man_file;
+  man_file.open(ProgramName() + "_man.rst");
```
junghans, line 67: I would be nice if the output file would not be hardcoded.

My thinking is something along the line of add an `--helpout` option, that specifies the file and could be used by `--man`, `--tex` and `--rst`.

@JensWehner any opinions?

JensWehner: Sure, in general I think that is sensible and makes it all more general. What would be a sensible default?

junghans: If no output file name is given, just print it to the screen (stdout).

junghans: In general printing to the screen might be enough as we can redirect all the output in CMake.

junghans: Did that.

junghans:
So I am kind of happy with it now.

However, I realized why we were doing the help to tex conversion with a script before. Some of our tools aren't C++ programs, but bash and python, hence they don't benefit from the `--man`, `--rst` option implemented in the application class. And if we need a script to convert output anyhow, we can as well do the same for the C++ tools.

@felipeZ @JensWehner: any opinions, what would be the cleanest here?

JensWehner:
So basically we have two options:

a) implement the `rst` option in python and bash as well
b) agree on one standard format and use a script to convert the rest. 

@felipeZ is the expert for python, but we are all on holiday now till the second week of january.

junghans:
This is not urgent, everything still works right now.

The problem with option a) is that we would need to add it to each and every script, which is quite cumbersome. For b) make the help output consistent isn't very hard and we already did that.

On a side note, we would need conversion to man format (groff) as well, there are tool like `help2man` to do that, but as far as I know they don't support rst output.

felipeZ:
> This is not urgent, everything still works right now.
> 
> The problem with option a) is that we would need to add it to each and every script, which is quite cumbersome. For b) make the help output consistent isn't very hard and we already did that.
> 
> On a side note, we would need conversion to man format (groff) as well, there are tool like `help2man` to do that, but as far as I know they don't support rst output.

@junghans  shall we then go for option b)? Converting between `rst` and `groff` is relatively easy in Python.

junghans:
Ok, let's abandon this PR then and go with a python script.

junghans:
I think the script should go into votca/tools, so that we can install into generate csg and xtp manpages in the module build as well (as we do for `votca_compare`).

PR #332, opened 2021-01-25
doc: make programs as subsections

junghans:
Improvement of https://www.votca.org/tools/reference.html

@JensWehner, should every program have its own html page instead of list all 3 tools in one page?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ Version 2022-dev
 ================
 
 -  create script to convert help to rst (#312, #328)
--  add rst pages to doc (#330)
+-  add rst pages to doc (#330, #332)
 -  remove tex and man output from app class (#329)
 
 Version 2021-rc.2 (released XX.01.21)
--- a/share/doc/index.rst
+++ b/share/doc/index.rst
@@ -2,7 +2,7 @@ TOOLS Manual
 ============
 
 .. toctree::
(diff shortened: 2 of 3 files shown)


JensWehner:
Good idea. I like this

JensWehner (approved):

junghans:
@jens, this isn't one page per-program yet, this PR just add headers.

If we want to do separate pages, we will need to reorg a bit more.

The problem is that the `votca_help2doc` generated pages doesn't include a header, hence sphinx will pick up the first section, i.e. `NAME`, for each page when adding the page with in `..toctree` instead of using `..include`.

JensWehner:
So we would have to automatically generate the headers as well? In that case I think it is better to leave them in one file.

junghans:
> So we would have to automatically generate the headers as well? In that case I think it is better to leave them in one file.

We could do that with `votca_help2doc` automatically.

Or create a simple rst file like:
```
PROGRAM
^^^^^^^^^^

..include PROGRAM.rst
```
for each executable.

felipeZ (approved):
LGTM

JensWehner:
@junghans I do not really care either way, I prefer the more maintainable solution.

junghans:
> @junghans I do not really care either way, I prefer the more maintainable solution.

Then this will the solution for now ;-)

PR #351, opened 2021-02-16
Joshua s brown/fix missing include

JoshuaSBrown:
std::sqrt is used in average.h but cmath is not included this fixes the problem.


JensWehner (approved):

JoshuaSBrown:
@junghans I don't know why the CI is failing in this case, seems unrelated to what I changed.

junghans:
Something in the notebook generation (@felipeZ will know better), I just retriggered it.

PR #354, opened 2021-02-25
Remove unused header

JoshuaSBrown:
Header unused and was causing problems for clang compilers as it was an extension specific to GNU compilers.


junghans:
This should go into stable, let's me rebase.

junghans (approved):

PR #360, opened 2021-04-23
added a NDim vector class

JensWehner:
This basically allows you to replace all the std::vector<std::vector<std::vector<T>> > things with a nicer interface, it should also be faster because it avoids a lot of pointer indrection for higher dimensions.

For numbers this should not be used because Eigen::Tensor is much better, but for other things which have to be stored on a regular grid for example this is nice.

### DIFF
--- a/include/votca/tools/NDimVector.h
+++ b/include/votca/tools/NDimVector.h
@@ -1,0 +1,104 @@
+/*
+ *            Copyright 2009-2020 The VOTCA Development Team
+ *                       (http://www.votca.org)
+ *
+ *      Licensed under the Apache License, Version 2.0 (the "License")
+ *
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *              http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#ifndef VOTCA_TOOLS_NDIMVECTOR_H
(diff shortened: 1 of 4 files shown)


JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: Added N-DimVector

rubengerritsen (changes requested):
Nice idea and it looks nice, I only see one problem, in the unit test what happens if you would ask for `a(0,1,30)`? There is no error checking in that case and it is quite likely that the linear index obtained from the inner product is still smaller than `storage_.size()`. Moreover if I try it with a very high number `a(100,100,100)` it happily returns 0 instead of a segmentation fault.

JensWehner:
So I can put assert statements in there, in release mode I want to be consistent with vector and also Eigen::matrix, which do not do any bounds checking.

include/votca/tools/NDimVector.h
```diff
@@ -0,0 +1,104 @@
...
+    static_assert((sizeof...(otherDimensions) + 1 == dim_),
+                  "Number of dimensions given does not match rank");
+    dimensions_ = {firstDimension, otherDimensions...};
+
+    Index size = std::accumulate(dimensions_.begin(), dimensions_.end(), 1,
+                                 std::multiplies<Index>());
+    storage_.resize(size);
+    offsets_[0] = 1;
+    std::partial_sum(dimensions_.begin(), dimensions_.end() - 1,
+                     offsets_.begin() + 1, std::multiplies<Index>());
```
felipeZ, line 63: Suddenly you became a functional programmer ;)

felipeZ (approved):
A very elegant solution!

rubengerritsen (approved):
Okay, I didn't release that not checking for bounds is default behavior, but than it is fine

include/votca/tools/NDimVector.h
JensWehner: yes,  I am starting to like this functional style. :)

JensWehner:
@votca-bot format

PR #361, opened 2021-04-30
Build fixes for F34 and Opensuse

junghans:


junghans:
@votca-bot changelog: fix build on F34 and openSUSE

PR #365, opened 2021-05-08
Propery extensions

JensWehner:
Made the string to type conversion much more powerful. Now `as<type>()` can be `bool` any numeric type or any kind of object that is constructible from a string and also a `std::vector<T>` of any of the above mentioned types, also Eigen::VectorXd Eigen::VectorXf and Eigen::Vector3d and Eigen::Vector3f are supported. 

Furthermore you can now delete Attributes and Children, although the later invalidates all pointers to child nodes. 

If we wanted we could go back to use a std::list underneath, instead of a std::vector and then all references would be valid, but then large xml files would take ages to load.

### DIFF
--- a/src/libtools/property.cc
+++ b/src/libtools/property.cc
@@ -16,10 +16,12 @@
  */
 
 // Standard includes
+#include <algorithm>
 #include <cstring>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
+#include <iterator>
 #include <stack>
 #include <stdexcept>
 #include <string>
@@ -50,7 +52,7 @@ const Property &Property::get(const string &key) const {
   }
 
   const Property *p;
-  map<string, Index>::const_iterator iter;
+  map<string, std::vector<Index>>::const_iterator iter;
   if (*n == "") {
     p = this;
   } else {
@@ -58,7 +60,7 @@ const Property &Property::get(const string &key) const {
     if (iter == _map.end()) {
       throw std::runtime_error("property not found: " + key);
     }
-    p = &_properties[iter->second];
+    p = &_properties[iter->second.back()];
   }
   ++n;
   try {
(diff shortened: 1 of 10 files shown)


junghans (approved):

JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: properties can be removed and string conversion much improved

src/libtools/property.cc
```diff
@@ -50,15 +52,15 @@ const Property &Property::get(const string &key) const {
   }
 
   const Property *p;
-  map<string, Index>::const_iterator iter;
+  map<string, std::vector<Index>>::const_iterator iter;
```
felipeZ, line 55: There is now a vector of indices in order to represent multiple attributes, right?

src/libtools/property.cc
```diff
@@ -73,6 +75,10 @@ const Property &Property::get(const string &key) const {
   return *p;
 }
 
+Property &Property::get(const string &key) {
+  return const_cast<Property &>(static_cast<const Property &>(*this).get(key));
```
felipeZ, line 79: Could you please briefly explain to me why is this double-casting needed? :)

src/libtools/property.cc
```diff
@@ -185,6 +187,37 @@ std::vector<Property *> Property::Select(const string &filter) {
...
+  if (indices.size() == 1) {
+    _map.erase(prop.name());
+  } else {
+    indices.erase(std::remove(indices.begin(), indices.end(), index_of_child),
+                  indices.end());
+  }
+
+  // if child is not the last element we have to do two things, a) switch the
+  // child with the last element and b) update the index of the former last
+  // element
```
felipeZ, line 208: Awesome comment here ;)

felipeZ (approved):

src/libtools/property.cc
JensWehner: yes to have indeces to all the properties which have the same tag.

src/libtools/property.cc
JensWehner: Basically it removes all the `const` , I think it was inspired by stack overflow.

JensWehner:
@votca-bot format

felipeZ (approved):

PR #366, opened 2021-05-10
Bool conv

JensWehner:
Blocked by https://github.com/votca/xtp/pull/704


felipeZ (approved):

CHANGELOG.rst
```diff
@@ -15,6 +15,7 @@ Version 2022-dev
 -  add missing cmath include (#351)
 -  removed eigenvalue algorithm, now done by eigen  (#352)
 -  Added N-DimVector (#360)
+-  properties can be removed and string conversion much improved (#365)
```
junghans, line 18: Huh? Wrong number?

junghans (commented):
Is this the same content as in #365?

CHANGELOG.rst
JensWehner: Yes this is a part of the PR which will be merged later, because some XTP functionality does not allow this change atm

junghans: Maybe add a line for #366 then as well.

junghans:
@felipeZ 

something is broken in xtp-tutorials:
```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-1-9f4d79d08fb3> in <module>
      3 import seaborn as sns
      4 columns = ["E(eV)", "epsGaussian","IM(eps)Gaussian",  "epsLorentz",  "Im(esp)Lorentz"]
----> 5 df = pd.read_table("methane_spectrum.dat", comment="#", sep='\s+',names=columns)
      6 sns.relplot(x="E(eV)", y="epsGaussian", ci=None, kind="line", data=df)
      7 plt.plot()
```

can you have a look?

felipeZ:
@junghans @JensWehner the xtp-tutorials is broken because there is an error in the calculation of the "methane_spectrum".  The error is raised in the step that generates the plot, but the problem is that not spectrum data is generated.

JensWehner:
yes, the reason for that is, that this change of bool screws up the whole option reading in xtp. That is the reason this is not merged yet.

PR #368, opened 2021-05-11
removed requirement for threading link library

JensWehner:
At the moment MKL is not found properly. 

We should in general rethink if we want to require MKL to be found and disable MKL with a cmake option instead of searching for it silently like atm, because the CI will not pick it up.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Version 2022-dev
 -  removed eigenvalue algorithm, now done by eigen  (#352)
 -  Added N-DimVector (#360)
 -  properties can be removed and string conversion much improved (#365)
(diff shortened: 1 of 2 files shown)


junghans:
Hmm I don't understand! The threading library is linked against and if not found it will fail at compile time.

junghans:
The threading library is part of the shared mil target:
https://github.com/votca/tools/blob/54e022204d1bbc
8a6a8aadfa51668e3d5610b86a/CMakeModules/FindMKL.cmake#L408

JensWehner:
Weird because it does not find it on my computer, but everything else runs fine then and we never looked for it before.

JensWehner:
CLosed because of #369

PR #371, opened 2021-05-13
cmake: mkl improvements

junghans:
@JensWehner remove https://github.com/votca/votca/blob/0aa9e0024c3ab94a365c60b6fae24399b3507155/CMakeLists.txt#L122_L124 and test locally


CMakeLists.txt
```diff
@@ -81,11 +81,26 @@ endif(Python_EXECUTABLE AND NOT PYTHON_INSTDIR)
...
-endif (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+set_package_properties(MKL PROPERTIES TYPE OPTIONAL PURPOSE "Enables accelerated performance with MKL")
+
+#openmp with clang smaller version 10 and user defined reductions is buggy https://bugs.llvm.org/show_bug.cgi?id=44134
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10.0)
+  # in case OpenMP was detected somewhere else
+  unset(OpenMP_FOUND)
+elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Intel" AND MKL_FOUND)
+  # part of mkl
+  unset(OpenMP_FOUND)
```
JensWehner, line 99: I do not get this part. So if MKL_FOUND then OpenMP is also found but we have to check its version is larger than 4.5, but MKL_FOUND can be true with gcc as well. Why do we unset(OpenMP_FOUND)?

JensWehner, line 99: I think this should go

JensWehner: because if we have gcc and mkl, we use gomp anyway. If we use MKL + icc we link aginst iopenmp which will be found again below

junghans: In the case of `gcc` or `clang`, `find_package(MKL)` will call `find_package(OpenMP)` as well, but not for version 4.5 so we have to detect it again.

For intel when using MKL we don't want to detect OpenMP, hence we unset `OpenMP_FOUND` just in case `find_package(OpenMP)` was called somewhere else (same for clang <=10) as we are using `OpenMP_FOUND` later to link `OpenMP::OpenMP_CXX`.

JensWehner: but do we not have  link against OpenMP::OpenMP_CXX even when using icc?

junghans: Didn't you say there was a conflict between the two or was that only true for gcc?

Anyhow, I dropped that case, let's we what breaks.

JensWehner (approved):
Thanks for all the work. :)

PR #372, opened 2021-05-27
removed unused option

JensWehner:


JensWehner:
@votca-bot format

src/libtools/application.cc
```diff
@@ -66,7 +66,6 @@ int Application::Exec(int argc, char **argv) {
     AddProgramOptions()("verbose1", "  be very loud and noisy");
     AddProgramOptions()("verbose2,v", "  be extremly loud and noisy");
     AddProgramOptions("Hidden")("man", "  output man-formatted manual pages");
```
junghans, line 68: Is the `man` option used at all?

JensWehner: I thought you use that for something.

junghans: No, we use that new `votca_help2doc` script everywhere now.

JensWehner: I thought that converts the man files to sth else.

junghans: It only uses the output of `--help`.

JensWehner: wow then we can remove quite a bit of stuff. :)

JensWehner:
@junghans can you review again

junghans (approved):

PR #374, opened 2021-06-17
add a add which can use . inside

JensWehner:
All a new add function which allows to add with `a.b.c` tag instead of just direct children


JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: added AddTree to property

junghans (approved):

PR #375, opened 2021-07-06
Options refactor

JensWehner:
Implementation of an optionsprocessor


JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: added optionshandler

CMakeModules/FindMKL.cmake
```diff
@@ -94,7 +94,6 @@ find_path(MKL_INCLUDE_DIR
   PATHS
     /opt/intel
     /opt/intel/mkl
-    /opt/intel/oneapi/mkl/latest
```
junghans, line 97: Why?

JensWehner: merge error, have reversed that.

include/votca/tools/tokenizer.h
```diff
@@ -53,8 +53,11 @@ inline T convert_impl(const std::string &s, type<T>) {
 inline bool convert_impl(const std::string &s, type<bool>) {
   if (s == "true" || s == "TRUE" || s == "1") {
     return true;
-  } else {
+
+  } else if (s == "false" || s == "FALSE" || s == "0") {
```
rubengerritsen, line 57: Maybe add a `tolower(s) == "false"` then a user can also type `False`

src/libtools/calculator.cc
```diff
@@ -18,151 +18,7 @@
 // Local VOTCA includes
```
rubengerritsen, line 18: Do we still need this file?

rubengerritsen (commented):
Looks good. 

I am wondering your `Link Resolver` now only acts on the defaults, if I am not mistaken. Could it be easily extended to also work on the user provided input file. 

I know we have had the discussion about the nested options files before and that you are not a fan. But for the qmmm calculations I am trying now, it makes much sense. If you create one dft and one gwbse option file that can be used for the vaccuum calculation, parameterization and the qmmm embedding that would be great, because as a user you do not need to worry about keeping basissets, functionals and convergence parameters in sync across many files. Also it seems that with the link resolver it could be just an optional feature instead of mandatory nesting.

src/libtools/calculator.cc
JensWehner: Not really, but on the other hand it is always good to not have to create it again and it does no harm, so I would leave it.

include/votca/tools/tokenizer.h
JensWehner: good point thanks

JensWehner:
Concerning having nested input files, yes technically it is no big problem, but it is not so easy, because we atm demand that options are inside the value of the xml for input files.

rubengerritsen (approved):

rubengerritsen:
Apparently my review is useless on this repo, but we can take up the nested options at a later time if we really want it

JensWehner:
@votca-bot format

rubengerritsen (approved):

JensWehner:
@votca-bot format

PR #377, opened 2021-07-27
factory can take other constructors now

JensWehner:
Made the factory more general to take also be able to construct objects which are not default constructible but need constructor arguments.


JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: extend factory

junghans (approved):

PR #378, opened 2021-08-05
Fix name from beadmass inference

JensWehner:


JensWehner:
@votca-bot changelog: changed tolerance for getting element names from mass

junghans:
Shouldn't that go into stable instead?

JensWehner:
I do not think it can go into stable because I also have to change CSG.

junghans:
> I do not think it can go into stable because I also have to change CSG.

Why not both?

JensWehner:
good point

junghans (approved):

PR #383, opened 2021-09-14
cmake: bring back REQUIRE_MKL

junghans:
This got lost in votca/votca@596b2c3daf65a1ebf24a924954d8f2b884fd54c5


JensWehner (commented):
I think when we look for mkl we should always use the iomp backend for openmp, because it should work with icpc, icpx and clang out of the box and it has a compability mode for gcc

If we do not find mkl just use the openmp backend we would use otherwise.

junghans:
> I think when we look for mkl we should always use the iomp backend for openmp, because it should work with icpc, icpx and clang out of the box and it has a compability mode for gcc

One can install mkl without `iomp5`, hence I don't wanted to force it.

> If we do not find mkl just use the openmp backend we would use otherwise.

I don't think `FindMKL.cmake` supports changing `MKL_THREAD_LAYER` on the fly.

All that said, in the end I really don't care much about `mkl`, just change it how you want it and merge.

JensWehner:
Good points.

JensWehner (approved):

